PR #4, opened 2023-02-15
Arm64 support

RichardoC:
Could you officially support ARM64 as well as AMD64?

This PR is currently working with the following exceptions

``` console

└─$ plz build //...                                                           
Build stopped after 1.49s. 2 targets failed:
    //examples/pipelines/scorecard-project/pipelinerun:pipelinerun
Error building target //examples/pipelines/scorecard-project/pipelinerun:pipelinerun: exit status 1
'GITHUB_TOKEN' is not set.

    //third_party/binary/tatskaari/go-deps:go-deps
1 error occurred:
        * Error retrieving https://github.com/Tatskaari/go-deps/releases/download/v1.4.2/go_deps_v1.4.2_linux_arm64: 404 Not Found

```
`scorecard` is because I don't have the required Github token and `go-deps` does not support ARM64 yet, though there is an [issue](https://github.com/Tatskaari/go-deps/issues/28) for it

I've not made the changes to enable ARM64 packages to be created, but think that would be the next step if you want to support this arch.

Thanks!

### DIFF
--- a/docs/getting-started/kind.md
+++ b/docs/getting-started/kind.md
@@ -93,7 +93,7 @@ We use [Kustomize Components](https://github.com/kubernetes-sigs/kustomize/blob/
 2. Run the following to create the Tekton Pipeline, Task, etc. resources on your cluster:
 
     ```bash
-    $ kustomize build | kubectl apply -f -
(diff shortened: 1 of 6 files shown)


northdpole:
we don't use please anymore, closing as go natively supports arm now

PR #20, opened 2023-03-10
Zap automation framework

northdpole:
this pull request modifies the zap producer to use [zap's automation framework](https://www.zaproxy.org/docs/desktop/addons/automation-framework/) instead of the old `zap-baseline.py` . The automation framework is more powerful and allows for better zap orchestration.
depends on #17 for fixing e2e tests


VJftw (approved):
Can merge when checks are passing 😀

PR #28, opened 2023-05-02
Go 123

### LINKED ISSUE #23: Fix existing lint issues
Running `plz lint -filter-mode=nofilter` results in 143 issues. The majority are fairly simple to fix but just time-consuming. We should be able to run the above without any results and make it the default.

northdpole:
closes #23


ptzianos:
In theory, this can be rebased on the latest main, so that we can get around to fixing the lint issues, however, reviewdog will probably force you to fix all the other lint issues in each of these files which is going to be quite a bit of work

northdpole:
closing as we do support go1.23

PR #53, opened 2023-12-04
Replace please with Makefile

### LINKED ISSUE #62: Replace please with Makefile
## Issue

Following discusision of #43 we have come to the realisation that `please` is not a build system that works for us due to the complexity it brings and the fact that v17 has refactored how a lot of the plugins and pleasings work which are crucial parts of our build process and they are now effectively broken. Furthermore, `please` is not used outside the context of `ThoughtMachine` and the majority of engineers out there would be unwilling to learn Tekton and Dracon AND please, in order to work with our system. We aim to reduce the barrier to entry and thus we need a build system that is simple, straightforward, widely known and well documented. That's why we decided to move to `make` instead which is available by default on all Unix-like systems and pretty much all engineers have used it in the past for some work. This should help reduce the learning curve of Dracon significantly and will allow anyone to be able to build, deploy and extend the system.

ptzianos:
This PR removes all please related configuration from the repository and replaces it with Makefile.

Changes include the following:
- [x] Remove all BUILD files*
- [x] Remove k3d documentation
- [x] Update KiND documentation**
- [x] Add documentation about all the components that require to be deployed
- [x] Cleanup example pipelines
- [x] Add files to allow users to deploy examples with helm
- [x] Cleanup openapi tests
- [x] Fix Github code quality checks
  - [x] test
  - [x] lint
  - [x] format
  - [x] test coverage
- [x] Add Github actions for producing releases based on tags
- [x] Allow user to define custom base images for components
- [x] Add draconctl to build pipelines and deprecate the kustomize-component-generator
- [x] Add draconctl subcommand for managing migrations

\* some BUILD files will remain in the repo, since they might be related to parts of the code that we need to discuss in the future if we want to support at all or if we want to completely get rid of
** KiND is the only officially supported deployment method at this point. users are going to be free to choose other methods of running their local clusters should they wish to. updated documentation will reflect this change  

Closes #62, #63, #64, #65, #66, #67, #68, #69, #70, #71, #72, #73, #74, #90, #101, #102, #103, #104, #105, #106, #107, #108, #109, #112, #118


northdpole:
great job pavlos!

northdpole (approved):

PR #60, opened 2024-02-03
Checkmarx producer

northdpole:
This component reads checkmarx output and serialises into Dracon format

Keep in mind that it does not orchestrate a remote checkmarx instance, instead it expects to find its input in the scratch directory

TODO:

- [ ] add OPTIONAL orchestration of remote Checkmarx


northdpole:
superseeded by #124

PR #85, opened 2024-02-14
add the ability to define arbitrary tags for scanInfo

northdpole:


northdpole:
closing since it has been folded into https://github.com/ocurity/dracon/pull/53

PR #88, opened 2024-02-16
Fix test output

northdpole:
this pr extends #62  by adding tests for all consumers who have the ability to write ScanInfo Tags


northdpole:
was merged as part of #53

PR #89, opened 2024-02-27
dependency-track parser

northdpole:


components/producers/dependency-track/main.go
```diff
@@ -0,0 +1,104 @@
...
+	issues := []*v1.Issue{}
+	for _, element := range *out {
+		iss := v1.Issue{}
+		target := element.Component.Purl
+		iss.Target = target
+		cwe := fmt.Sprintf("%d", element.Vulnerability.CweID)
+		iss.Type = element.Vulnerability.VulnID
+		iss.Title = element.Vulnerability.Title
+		iss.Cvss = element.Vulnerability.CvssV3BaseScore
+		iss.Severity = v1.Severity(v1.Severity_value[fmt.Sprintf("SEVERITY_%s", element.Vulnerability.Severity)])
```
nscuro, line 49: I suggest doing a `switch` on severity instead. It's possible (although rare) that a vulnerability has no severity assigned yet, in which case you'd get an `UNASSIGNED` here.

The `v1.Severity` model doesn't cover that.

See: https://github.com/DependencyTrack/hyades-apiserver/blob/b1c3b0d53f7f3db01f7d3d210e25646460e1d765/src/main/java/org/dependencytrack/model/Severity.java#L30-L35

components/producers/dependency-track/main.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func parseIssues(out *DependencyTrackOut) ([]*v1.Issue, error) {
+	issues := []*v1.Issue{}
+	for _, element := range *out {
+		iss := v1.Issue{}
+		target := element.Component.Purl
+		iss.Target = target
+		cwe := fmt.Sprintf("%d", element.Vulnerability.CweID)
```
nscuro, line 45: Same as below, it'd be better to rely on the `cwes` array instead.

PR #92, opened 2024-02-28
blackduck parser

northdpole:


components/producers/blackduck/main.go
```diff
@@ -0,0 +1,204 @@
...
+			continue
+		}
+		res = append(res, int32(newIntID))
+	}
+	return res
+}
+
+func BDTargetToPurl(originName, originID string) string {
+	splitOrigin := strings.Split(originID, ":")
+	return packageurl.NewPackageURL(originName, splitOrigin[0], splitOrigin[1], splitOrigin[2], packageurl.Qualifiers{}, "").ToString()
```
nscuro, line 59: Unfortunately the origin name doesn't always map 1:1 with the [recognized PURL types](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst). I am unable to locate BD docs about what origins it may report, but [this KB article](https://community.synopsys.com/s/article/What-is-an-Origin-and-Origin-ID-in-Blackduck) already tells us that NPM packages will be reported with origin `npmjs`, whereas the correct PURL type is `npm`.

Consider adding a sanity-check here that logs a warning or so when `originName` isn't found in either [`KnownTypes` or `CandidateTypes`](https://pkg.go.dev/github.com/package-url/packageurl-go#pkg-variables). That should allow you to discover irregularities and add mappings over time.

components/producers/blackduck/main.go
```diff
@@ -0,0 +1,204 @@
...
+			log.Println(err)
+			continue
+		}
+		res = append(res, int32(newIntID))
+	}
+	return res
+}
+
+func BDTargetToPurl(originName, originID string) string {
+	splitOrigin := strings.Split(originID, ":")
```
nscuro, line 58: Note that some ecosystems don't have a concept of namespaces (Python, occasionally NPM, Rust, ...), so you might only get 2 parts in return for this `Split`.

components/producers/blackduck/main.go
nscuro: Non-exhaustive list of known mappings:

| BD Origin Name | PURL |
|:------------|:--------------|
| `anaconda` | [`pkg:conda/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#conda) |
| `centos` | [`pkg:rpm/centos/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm) |
| `debian` | [`pkg:deb/debian/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#deb) |
| `fedora` | [`pkg:rpm/fedora/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm) | 
| `gradle` | [`pkg:maven/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#maven) |
| `npmjs` | [`pkg:npm/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#npm) |
| `opensuse` | [`pkg:rpm/opensuse/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm) |
| `packagist` | [`pkg:composer/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#composer) |
| `redhat` | [`pkg:rpm/redhat/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm) |
| `unknown` | [`pkg:generic/...`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic) |

components/producers/blackduck/main.go
northdpole: you mean something like this?

components/producers/blackduck/main.go
northdpole: can you please take another look?

nscuro (approved):

PR #125, opened 2024-03-18
flatten dracon scan library methods

northdpole:


ptzianos (changes requested):
This can be simplified a lot by doing the following:

```golang
type flattened struct {
	ToolName      string                 `json:"tool_name,omitempty"`
	ScanUuid      string                 `json:"scan_uuid,omitempty"`
	ScanStartTime *timestamppb.Timestamp `json:"scan_start_time,omitempty"`
	*Issue
}

func Marshal(toolName string, si ScanInfo, issue *Issue) ([]byte, error) {
	modifiedMap := map[string]string{}
	for k, v := range si.ScanTags {
		modifiedMap["tag-"+k] = v
	}
	modifiedMapBytes, err := json.Marshal(modifiedMap)
        if err != nil {
                return nil, err
        }

        flattenedIssue := flattened{
                toolName,
                si.ScanUuid,
                si.ScanStartTime,
                issue,
        }
        marshaledFlattenedIssue, err := json.Marshal(flattenedIssue)
        if err != nil {
                return nil, err
        }
        marshaledFlattenedIssue[len(marshaledFlattenedIssue)-1] = ','
        marshaledFlattenedIssue = append(marshaledFlattenedIssue, modifiedMapBytes[1:]...)
        return marshaledFlattenedIssue, nil
}
```

ptzianos:
Closing as not relevant any more

Issue #144, opened 2024-04-25
Document how to use Helm packages for Tekton Tasks in a more efficient way


ptzianos:
We need a better way of managing Tekton Tasks and Pipelines and automatically discovering components available for the system to use without having to manually compile the \`componentsConfig\` and to allow the open source system to have multiple Helm managed Pipelines deployed at the same time in the same namespace. Document how we can achieve that with a reasonable amount of detail and hopefully a small amount of effort

ptzianos:
merged as part of https://github.com/ocurity/dracon/pull/145

PR #145, opened 2024-04-25
#144 RFC - Improve integration and Helm workflow for Tasks

ptzianos:


docs/dep/dep-004.md
```diff
@@ -0,0 +1,76 @@
...
+administrator of the system to make sure that the Task is in a state ready to be deployed.
+
+When a user wishes to refer to a Task deployed via Helm on the cluster, they will have to use a
+new URI type that is inspired by `pURL` and looks like this:
+`helm://{{ helm_package_id }}/name`
+
+The `helm_package_id` is the name of the Helm release deployed on the cluster and will be used to
+list the configmaps of the namespace and the Tasks using the label with name
+`app.kubernetes.io/instance` and value the `helm_package_id`. The system will expect to find a
+configmap with the label `v1.dracon.ocurity.com/bill-of-materials: tekton-tasks`. The configmap
```
northdpole, line 49: i don't understand why a configmap makes sense in this particular scenario of open source components?
shouldn't we instead have either `draconctl` listing the tasks installed in a cluster or the available tasks in a repository or OCI registry?

ptzianos: I've significantly updated the RFC and removed this part after our offline discussion to simplify the structure of the system.

ptzianos:
Stacked on top of https://github.com/ocurity/dracon/pull/156

docs/dep/dep-004.md
```diff
@@ -0,0 +1,146 @@
...
+deployed on the cluster similar to the case where we have a file in the local
+filesystem.
+
+3. a pURL-inspired id of a Task deployed using Helm
+
+As mentioned previously, we wish to package all our Tasks in one Helm package
+that can be easily deployed on a cluster and provides access to all our
+components. When a user wants to reference a Task deployed via Helm on the
+cluster in a Pipeline description, they will have to use a new URI type that is
+inspired by `pURL` and looks like this: `helm://{{ helm_release_name }}/name`.
```
northdpole, line 70: nit: 
why not make it `purl://helm/{{ helm_release_name }}/{{name}}` ? then we can use the go-purl library to parse it and we don't need to do any custom magic

docs/dep/dep-004.md
```diff
@@ -0,0 +1,146 @@
...
+type Applier struct {
+    Apply()
+}
+
+type Component struct {
+    ID() string
+
+    TektonV1Beta1Task() (*tektonv1beta1api.Task, error)
+
+    ManagedByHelm() bool
```
northdpole, line 95: i think ALL components should be managed by helm, this simplifies things and allows us to have all components as helm packages

docs/dep/dep-004.md
```diff
@@ -0,0 +1,146 @@
...
+files, they will be applied on the cluster. If no error is returned the client
+can invoke the `Deploy` method which will deploy the Pipeline itself on the
+cluster.
+
+It is important to note, that the Orchestrator does not manage the assembling
+of a Pipeline or the validation. It expects to receive a correct object and
+will try to deploy whatever pipeline isprovided.
+
+The `Component` struct is an abstraction layer on top of any component we are
+using. It will internally resolve the Task URI and figure out how to manage it,
```
northdpole, line 117: > It will internally resolve the Task URI and figure out how to manage it,
If we make everything be a HELM task, then we don't need multiple managers.
Local filesystem management is only useful for development and I'd say, maybe even relatively useless since we can build and load charts easily on local registries.

Furthermore, having the ability to treat everything as a "remote" chart, allows us  to move away from Tekton task defs and support a number of execution backends if we want to

docs/dep/dep-004.md
ptzianos: :thinking: I think you are right that I'm not following purl closely but what you are suggesting is a bit closer but not enough. looking at the [examples](https://github.com/package-url/purl-spec?tab=readme-ov-file#some-purl-examples) once again I think the closest we could go is: `pkg:helm/{{ helm_release_name }}/name` or we could go further and do `pkg:helm/{{ helm_registry }}/{{ package_name }}/name` so that we can be more explicit. WDYT?
(on a side note, this could be an interesting contribution to the pURL spec itself)

docs/dep/dep-004.md
ptzianos: I added this so that a user can have an easy way to deploy a Task from the local repo. suppose that you are hacking away trying to get something to work. if you have to package it in Helm and deploy it that adds a couple of steps. allowing ppl to just fetch files from the filesystem or from a URL like we do now should make development a smoother process

docs/dep/dep-004.md
ptzianos: I don't see how Helm has anything to with which backend we are using. we could be using Helm to package Tekton V1Beta1 or V1 Tasks, custom deployments and CRDs or whatever else we want. Helm doesn't know or care about the type of object, it just knows that it's some YAML that potentially the K8s API master will recognise with probably some Go templating in it. For me the question of allowing extra ways of fetching a Task is more about development experience as I mentioned in a previous comment

docs/dep/dep-004.md
northdpole: yes, absolutely, my aim was to do a valid purl so let's do `pkg:helm/{{ helm_registry }}/{{ package_name }}/name`
where for us it is ALWAYS `pkg:helm/{{ helm_registry }}/dracon/{{ package_name }}` so we can query our components

docs/dep/dep-004.md
northdpole: sure, then makes sense, trying to keep it simple but since helm allows for the installation of local packages why don't we put both components through a `helm upgrade` but in the case of a purl we resolve the purl first?

docs/dep/dep-004.md
northdpole: sure, makes sense, then please disregard my comment

northdpole (approved):

docs/dep/dep-004.md
ptzianos: So, I created a toy example to see how the purl library would treat the different format and apparently both of the work.
```go
package main

import (
	"fmt"

	packageurl "github.com/package-url/packageurl-go"
)

func main() {
	taskPurl, err := packageurl.FromString("pkg:helm/helm.github.com/dracon/consumer-aggregator@1.1.0")
	fmt.Printf("%v\n%v", taskPurl, err)
}
```
In this case the `Namespace` will be the `helm.github.com/dracon` part.

I think for now we can use the simple version of the purl `pkg:helm/{{ helm_release_name }}/name` where the `Namespace` will be the Helm release name and will be used as a label selector as described. In the future, if we decide to expand our integration with Helm we could allow `Namespaces` that contain the host of the Helm registry and the explicit package from it, but for the time being I think it would too much complexity for no good reason.

docs/dep/dep-004.md
ptzianos: we would need to also build the local packages and make sure that the version of the local chart is greater than the one deployed so that the `upgrade` command can behave as we expect. that's why I think it's a bit fussy and better left for the future if needed

PR #151, opened 2024-04-25
[OCU-90] Make Dracon-dev into one Helm Package

northdpole:
Implements #150


ptzianos (changes requested):
Please split away the RFC part into a separate PR, we first review and merge that, then we can proceed to reviewing this, but only after it's working and the commit history has been cleaned up

northdpole:
> Please split away the RFC part into a separate PR, we first review and merge that, then we can proceed to reviewing this, but only after it's working and the commit history has been cleaned up

thanks for the review, done

Makefile
```diff
@@ -239,4 +208,24 @@ deploy-tektoncd-dashboard: tektoncd-dashboard-helm
 		--values ./deploy/tektoncd/dashboard/values.yaml \
 		--namespace $(TEKTON_NS)
 
-dev-deploy: deploy-nginx deploy-arangodb deploy-kibana deploy-mongodb deploy-pg deploy-tektoncd-pipeline deploy-tektoncd-dashboard
+deploy-dracon-dev: deploy-elasticoperator
+	make draconctl-image-publish CONTAINER_REPO=localhost:5000/ocurity/dracon
```
ptzianos, line 212: I don't think we need this, the user can run this command themselves if needed

Makefile
```diff
@@ -239,4 +208,24 @@ deploy-tektoncd-dashboard: tektoncd-dashboard-helm
...
+
+dev-deploy: 
+	make deploy-nginx
+	make deploy-arangodb
+	make deploy-tektoncd-pipeline
+	make deploy-tektoncd-dashboard
+	make deploy-dracon-dev
+
+dev-teardown:
+	kind delete clusters dracon-demo
```
ptzianos, line 229: forgot an `@` over here

Makefile
```diff
@@ -239,4 +208,24 @@ deploy-tektoncd-dashboard: tektoncd-dashboard-helm
...
+	make draconctl-image-publish CONTAINER_REPO=localhost:5000/ocurity/dracon
+	@helm upgrade dracon ./deploy/dracon/ \
+	 	  --install \
+		  --values ./deploy/dracon/values.dev.yaml \
+		  --create-namespace \
+		  --namespace $(DRACON_NS) \
+		  --set "global.image.draconVersion=$(DRACON_VERSION)"
+		  --wait
+
+dev-deploy:
```
ptzianos, line 221: you can simplify this to `dev-deploy: deploy-nginx deploy-arangodb deploy-tektoncd-pipeline deploy-tektoncd-dashboard deploy-dracon-dev`

deploy/dracon/charts/elasticsearch-0.1.0.tgz
ptzianos, line 1: I think we should add this folder to the `.gitignore` these will be pulled when the help install/upgrade command is executed

deploy/dracon/templates/job.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: batch/v1
```
ptzianos, line 1: i think we should add a boolean to enable migrations and have the entire job in an if so that if for whatever reason the migrations are not supposed to be deployed, they can be skipped

ptzianos (changes requested):
5 nits and we are good to go :rocket:

PR #157, opened 2024-04-29
helmify rfc

northdpole:


docs/designs/helmify.md
```diff
@@ -0,0 +1,44 @@
...
+
+It is suggested that the project uses Helm as its sole way of deploying things.
+
+## Changes suggested
+
+The following changes are suggested
+
+* Make a `dracon-dev` chart which installs all the sub-charts referenced by 
+the `make-dev` target
+* Make `draconctl migrations apply` only apply migrations against a database url
```
ptzianos, line 20: This doesn't make sense. The `draconctl migrations apply` always connects to a database using a URI and applies a set of migrations. Whether that happens while on an admins machine or inside a pod is irrelevant. The whole point of this work to not have a separate manual stage for running the command that deploys itself as a pod but to have it as an automated part of the Helm package. And why should it not be a job deployed by Helm? Helm can monitor jobs and track their progress.

docs/designs/helmify.md
```diff
@@ -0,0 +1,44 @@
...
+and not try to manage a k8s job
+* Make a `dracon` chart which installs dracon components and migrations only
+assuming there are prerequisite systems present (Elasticsearch, Postgres etc),
+values for those systems are provided through `values.yaml`
+* Transform all components into Charts, one chart for each.
+* Make `pipelines` be their own Charts
```
ptzianos, line 26: this is already the case, the other [RFC](https://github.com/ocurity/dracon/pull/145) implementation will tackle removing the Tasks from the Pipeline definitions

docs/designs/helmify.md
```diff
@@ -0,0 +1,44 @@
...
+The following changes are suggested
+
+* Make a `dracon-dev` chart which installs all the sub-charts referenced by 
+the `make-dev` target
+* Make `draconctl migrations apply` only apply migrations against a database url
+and not try to manage a k8s job
+* Make a `dracon` chart which installs dracon components and migrations only
+assuming there are prerequisite systems present (Elasticsearch, Postgres etc),
+values for those systems are provided through `values.yaml`
+* Transform all components into Charts, one chart for each.
```
ptzianos, line 25: We have discussed this offline that it's not necessary at this stage, we might revisit it in the future and even if it was, it's not part of the work this RFC is supposed to tackle

docs/designs/helmify.md
```diff
@@ -0,0 +1,44 @@
...
+ yamls applied through `draconctl` (migrations). This is significantly confusing
+and tech debt since it is not clear what installs what, how and why.
+
+It is suggested that the project uses Helm as its sole way of deploying things.
+
+## Changes suggested
+
+The following changes are suggested
+
+* Make a `dracon-dev` chart which installs all the sub-charts referenced by
```
ptzianos, line 18: there is no `make-dev` target, there is a `dev-deploy` target

docs/designs/helmify.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Changes suggested
+
+The following changes are suggested
+
+* Make a `dracon-dev` chart which installs all the sub-charts referenced by 
+the `make-dev` target
+* Make `draconctl migrations apply` only apply migrations against a database url
+and not try to manage a k8s job
+* Make a `dracon` chart which installs dracon components and migrations only
```
ptzianos, line 22: The migrations should be applied by the chart that deploys the database where the migrations will be applied

northdpole: there is no individual database chart. Either we install postgres as part of the dracon-dev values.dev.yaml or we just run the migrations as a dracon post-install job against the database string identified in the values.yaml

docs/designs/helmify.md
```diff
@@ -0,0 +1,44 @@
...
+assuming there are prerequisite systems present (Elasticsearch, Postgres etc),
+values for those systems are provided through `values.yaml`
+* Transform all components into Charts, one chart for each.
+* Make `pipelines` be their own Charts
+* Add the ability for draconctl to interface with an OCI registry and discover 
+  components.
+* Add the ability for draconctl to create Charts via CLI arguments and manage
```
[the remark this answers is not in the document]
northdpole: a train of thought since abandoned, removed it

docs/designs/helmify.md
```diff
@@ -0,0 +1,44 @@
+# Helm-ify
+
```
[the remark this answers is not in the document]
northdpole: A thing with Tekton:
It's a mess, there's tons of vars they need to make configurable beyond the basics. There's a very good reason they haven't created one themselves yet. I don't wanna pile up more work for us let alone more work we need to maintain for free without any benefit. Let's keep it in the repo for now and if we ever feel generous enough let's create an external chart

ptzianos (approved):

PR #161, opened 2024-05-01
Add new K8s clientset and replicate kubectl apply behaviour

ptzianos:
Adds new clientset that is a combination of the Tekton V1Beta1 clientset along with a fake clientset to allow users to mock calls to the K8s API efficiently. The migrations commands are modified to use the new `Apply` method of the K8s clientset that replicates the `kubectl apply` behaviour


ptzianos:
The commits that are missing from the history are completely irrelevant to this, so I don't expect any conflicts, the only PR that is using this code is stacked on top of this and it's in draft form atm

PR #162, opened 2024-05-01
Refactor migrations commands to prevent subtle concurrency bugs

ptzianos:
stacked on top of: https://github.com/ocurity/dracon/pull/161

* 6dd5f567dd8d6d1446d36c8c3d1914d62711d486 - fixes 159: modify draconctl compiling make target to always build binary
    
    The previous version of the draconctl compiling make target would only
    compile the binary if the binary didn't exist in the expected path. This
    however is a problem if the code changes, because Makefile doesn't have
    the ability to track that and recompile the binary. This commit changes
    the target to be a phony one so that the binary is always recompiled if
    necessary. Go internally is able to track changes and produce a new
    binary only when needed, so this shouldn't cause anything to slow down
    but it will definitely improve the development process.

* 1e01522bc39dea5c9662d54b7c8f78e272f66510 - fixes 153: refactor 'draconctl migrations apply' command
    
    The previous version of the 'migrations apply' command suffered from a
    couple of shortcomings:
    
    1. The code tracking the logs and success of the generated pods was
       fairly complex and monolithic
    2. there were no tests for the pod tracking code
    3. if the code polling for new pods found one that had succeeded it
       would not show in the output the logs of the pod and it would exit
       immediately which could cause a user to misunderstand what was
       happening.
    4. If a SIGTERM signal was sent to the process inside the pod, the
       leader election loop was not terminated cleanly
    5. there was no proper error management since the last error would be
       returned which could be the signal handler having its channel closed
       or the leader election failure hook being invoked because the success
       of the command triggered the code to release the lock.
    
    All the above issues are fixed with this commit that does the following:
    1. the code tracking the progress of the pods is split away into the
       `pkg/k8s` package
    2. the tracking of the pod logs is split away from the pod polling
       functionality and the job watching functionality to make it easier to
       test.
    3. the leader election loop has been modified to use the sync.Once
       object to make sure that only the first error that triggers the exit
       of the main loop is recorded, so that we can get a better idea of
       what happened.
    5. a timeout parameter is introduced and a more careful use of contexts
       with timeouts is used to ensure that the code gracefully shuts down.


andream16:
It looks very out of date - feel free to re-open it if relevant

PR #166, opened 2024-05-01
[OCU-96] dracon component defs

northdpole:


ptzianos (changes requested):
I still don't understand what we are trying to solve over here

PR #173, opened 2024-05-10
Fix Python Workflow

northdpole:
Closes 
* [OCU-103] by removing the custom `*-tool` containers from safety and bandit (tools exist in pypi we can easily install on runtimee.
* [OCU-102] by slightly updating the custom package deployment instructions


northdpole:
@ptzianos seems that `npm run format` is broken, it consistently claims to have rewritten all md files erroneously (lint and git show nothing)

flowirtz (approved):

ptzianos:
> @ptzianos seems that `npm run format` is broken, it consistently claims to have rewritten all md files erroneously (lint and git show nothing)

Yes, the markdown formatter has this annoying habit of printing this message even though it hasn't changed anything in the file. This is a bug but the formatter works as expected. I checked out the branch and I ran the `make fmt` command and it did produce some minor fixes:

```markdown
pavlos@pavlos-framework:~/projects/ocurity/dracon-oss$ git diff
diff --git a/docs/getting-started.md b/docs/getting-started.md
index 2e61320a..612d28ab 100644
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -195,7 +195,6 @@ that will be picked up by the Makefile and build the container.
 You can package your components into a Helm package by running the following
 command:
 
-
 ```bash
 export CUSTOM_DRACON_VERSION=$(make print-DRACON_VERSION)
 export CUSTOM_HELM_COMPONENT_PACKAGE_NAME=

...
```

The fmt command doesn't check the stdout it only checks if some change has been applied to the files

PR #181, opened 2024-05-14
Ensure dracon_scan_* variables are set correctly

### LINKED ISSUE #180: Env variables for scan tags/id/start_time are not added to producers
Tekton is the gift that keeps on giving. In our local test environment we are using Tekton v0.44.0. Turns out that this version of Tekton is looser when it comes to parameter validation because it fails to see that we are setting the parameters for the `dracon-scan-id`, `dracon-scan-tags`  and `dracon-scan-start-time` but these parameters are not present in any of the producer Tasks packages in our Helm packages. However, Tekton v0.40.0, which is the version we use in some of our private test environments can identify this issue. This doesn't prevent our Tasks from running (which should be investigated separately too).

We should add the expected parameters to the Producer Tasks and we should regenerate our Helm packages, consolidate the code that processes each Task so that it can be re-used in Tekton Orchestrator which could deploy Tasks too and add tests that explicitly check this.

ptzianos:
Different versions of Tekton apply different validations to a Pipeline. If one uses more than version they will probably get a more complete list of errors for their pipelines. After using different versions of Tekton, I discovered that the `dracon_scan_*` parameters for the producer tasks are not set correctly.

This PR fixes two issues:
1. the `dracon_scan_*` parameters are set correctly
2. the code that adds anchors and the `dracon_scan_*` parameters is consolidated in one place and tests are added to make sure that the Tasks are processed before being packaged in a Helm package or before getting deployed on the fly by the orchestrator.

fixes #180

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -78,7 +78,7 @@ Kibana and the bitnami charts to deploy instances of PostgreSQL and MongoDB.
 If you run the command:
 
 ```bash
-make dev-dracon DRACON_VERSION=v0.8.0
+make dev-dracon DRACON_VERSION=v0.13.0
 ```
 
 You will deploy the Elastic Operator on the cluster and the Dracon Helm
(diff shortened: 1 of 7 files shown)


ptzianos:
All lint errors have been fixed, the linter is erroneously still showing them as errors though

PR #185, opened 2024-05-15
update the publish workflow to also build and publish dracon itself a…

### LINKED ISSUE #186: Make Publish job also publish draconctl and dracon itself

northdpole:
…nd draconctl

closes #186

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,14 +24,21 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Publish Docker images
         run: make publish-containers
-
+      - name: Publish Draconctl image
+        run: make draconctl-image-publish
+      - name: Setup Go
+        uses: actions/setup-go@v4
       - name: Publish component Helm package
         run: |
-          set -e;
-          DRACON_VERSION_SEMVER=$(sed 's/v//' <<< ${{ github.ref_name }});
-          make cmd/draconctl/bin; \
-          bin/cmd/draconctl components package --version ${{ github.ref_name }} --chart-version ${DRACON_VERSION_SEMVER} --name dracon-oss-components ./components; \
+          set -e
+          DRACON_VERSION_SEMVER=$(sed 's/v//' <<< ${{ github.ref_name }})
+
+          go install github.com/ocurity/dracon/cmd/draconctl@latest
+          ~/go/bin/draconctl components package --version ${{ github.ref_name }} --chart-version ${DRACON_VERSION_SEMVER} --name dracon-oss-components ./components;
+
+          helm package --version=${DRACON_VERSION_SEMVER} --app-version=${{ github.ref_name }} deploy/dracon/chart
+
           helm push dracon-oss-components-${DRACON_VERSION_SEMVER}.tgz oci://ghcr.io/ocurity/dracon/charts
+          helm push dracon-${DRACON_VERSION_SEMVER}.tgz oci://ghcr.io/ocurity/dracon/charts
--- a/Makefile
+++ b/Makefile
@@ -213,13 +213,12 @@ dev-dracon: deploy-elasticoperator deploy-arangodb-crds add-bitnami-repo
 		  --values ./deploy/dracon/values/dev.yaml \
 		  --create-namespace \
 		  --namespace $(DRACON_NS)\
-		  --version $$(echo "${DRACON_VERSION}" | sed 's/^v//')\
+		  --set "enrichmentDB.migrations.suffix=$$(echo "${DRACON_VERSION}" | sed 's/\./-/g')"\
 		  --set "enrichmentDB.migrations.image=$(CONTAINER_REPO)/draconctl:$(DRACON_VERSION)"\
 		  --wait
 	@helm upgrade dracon-oss-components oci://ghcr.io/ocurity/dracon/charts/dracon-oss-components \
 		--install \
 		--namespace $(DRACON_NS) \
-		--version $$(echo "${DRACON_VERSION}" | sed 's/^v//')
 
 dev-infra: deploy-nginx deploy-tektoncd-pipeline deploy-tektoncd-dashboard
 
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,8 @@ draconctl-image: cmd/draconctl/bin
 
 draconctl-image-publish: draconctl-image
 	$(DOCKER) push "${CONTAINER_REPO}/draconctl:${DRACON_VERSION}"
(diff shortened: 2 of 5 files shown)


deploy/dracon/chart/Chart.yaml
```diff
@@ -4,7 +4,6 @@ description: |
   A Helm chart for Kubernetes containing services needed for Dracon pipelines to run. Please check the documentation for more information
 type: application
 version: 0.7.0
```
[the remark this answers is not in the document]
northdpole: every helm chart needs a version, can't remove it

.github/workflows/publish.yml
```diff
@@ -27,11 +27,18 @@ jobs:
...
+      - name: Setup Go
+        uses: actions/setup-go@v4
       - name: Publish component Helm package
         run: |
           set -e;
           DRACON_VERSION_SEMVER=$(sed 's/v//' <<< ${{ github.ref_name }});
-          make cmd/draconctl/bin; \
-          bin/cmd/draconctl components package --version ${{ github.ref_name }} --chart-version ${DRACON_VERSION_SEMVER} --name dracon-oss-components ./components; \
-          helm push dracon-oss-components-${DRACON_VERSION_SEMVER}.tgz oci://ghcr.io/ocurity/dracon/charts
+          go install github.com/ocurity/dracon/cmd/draconctl@latest
```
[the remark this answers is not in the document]
northdpole: fixed, no idea why it worked before, but better to explicitly setup go i think, more robust

Makefile
```diff
@@ -56,6 +56,8 @@ draconctl-image: cmd/draconctl/bin
 
 draconctl-image-publish: draconctl-image
 	$(DOCKER) push "${CONTAINER_REPO}/draconctl:${DRACON_VERSION}"
+	$(DOCKER) tag "${CONTAINER_REPO}/draconctl:${DRACON_VERSION}" "${CONTAINER_REPO}/draconctl:latest"
```
northdpole, line 59: @ptzianos  please take a look,
Due to rapid development i think having the ability to use this image from the latest tag always makes sesne

ptzianos: if we publish it as the latest, we push the `v0.14.0` tag, it's published, then we trigger for some reason the v0.13.0 publish job because it failed and now we have a wrong reference to latest. I think we should push with both tags, so that at least we can pin stuff to the version where needed and play looser wherever we want

.github/workflows/publish.yml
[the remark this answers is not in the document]
ptzianos: yeah I think we got away with it by sheer luck but it's probably going to break unless we define it explicitly

deploy/dracon/chart/Chart.yaml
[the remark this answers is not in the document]
ptzianos: you can pass the version as a parameter when building the chart same way you do for the appVersion

Makefile
northdpole: that's exactly what it's doing, look one line below :D

.github/workflows/publish.yml
```diff
@@ -24,14 +24,21 @@ jobs:
...
       - name: Publish component Helm package
         run: |
-          set -e;
-          DRACON_VERSION_SEMVER=$(sed 's/v//' <<< ${{ github.ref_name }});
-          make cmd/draconctl/bin; \
-          bin/cmd/draconctl components package --version ${{ github.ref_name }} --chart-version ${DRACON_VERSION_SEMVER} --name dracon-oss-components ./components; \
+          set -e
+          DRACON_VERSION_SEMVER=$(sed 's/v//' <<< ${{ github.ref_name }})
+
+          go install github.com/ocurity/dracon/cmd/draconctl@latest
```
ptzianos, line 38: why do this? you already checked out the latest version of the codebase and you are executing commands in it

.github/workflows/publish.yml
```diff
@@ -24,14 +24,21 @@ jobs:
...
+          ~/go/bin/draconctl components package --version ${{ github.ref_name }} --chart-version ${DRACON_VERSION_SEMVER} --name dracon-oss-components ./components;
+
+          helm package --version=${DRACON_VERSION_SEMVER} --app-version=${DRACON_VERSION_SEMVER} deploy/dracon/chart
```
ptzianos, line 41: the `--app-version` should be set to `${{ github.ref_name }}` in the same way as it is in the previous command so that we can reference the original tag of the git repo

Makefile
```diff
@@ -211,6 +213,7 @@ dev-dracon: deploy-elasticoperator deploy-arangodb-crds add-bitnami-repo
 		  --values ./deploy/dracon/values/dev.yaml \
 		  --create-namespace \
 		  --namespace $(DRACON_NS) \
+          --version $$(echo "${DRACON_VERSION}" | sed 's/^v//')
```
ptzianos, line 216: the line has a different format than the rest of the command and it's missing a slash at the end. there is a small bug at the next line that's also missing a slash at the end, unrelated to this PR but I think we should fix it since we are working here.

deploy/dracon/chart/values.yaml
```diff
@@ -64,7 +64,14 @@ enrichmentDB:
   # enrichment database
   connectionStr: ""
 
+dracon-components:
```
ptzianos, line 67: I don't see any change in the Helm manifests where this is used. why is it here?

deploy/dracon/chart/values.yaml
```diff
@@ -64,7 +64,14 @@ enrichmentDB:
...
+  # if you're developing components
+  #  set this to false to install your own components
+  enabled: true
+
 global:
   image:
     # registry to use for all 
-    registry: ""
+    registry: "ghcr.io"
```
ptzianos, line 75: all our images already have a default value for the registry in the Helm template, I don't think we need this here or if we put it here, we should remove the use of the `default` function in the Helm templates

deploy/dracon/chart/Chart.yaml
```diff
@@ -13,4 +11,4 @@ dependencies:
   - name: mongodb
     version: 15.1.5
     repository: https://charts.bitnami.com/bitnami
-    condition: mongodb.enabled
+    condition: mongodb.enabled
```
ptzianos, line 14: why remove the last line?

.github/workflows/publish.yml
northdpole: it's a matter of principle:
in this scope draconctl is a tool used to package and publish things,  hence we use the latest stable version

if we were to be testing draconctl then we can build it from the binary
i don't see how this is a problem

ptzianos: you are not using the latest stable version, you are using the latest commit in the main branch. these are not the same unfortunately

PR #190, opened 2024-06-03
add structured dracon logging

northdpole:
This PR leverages Go's `log/slog` package to add a default logger to every logging statement we have in the project.
This allows us to change logging for all components easily and in the future be more flexible with smarter things like tracing or making groups out of our logs.

Enrichers do not have a base method for now so they don't get structured logging, this is coming in the next PR


cmd/draconctl/main.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"log/slog"
 	"os"
 
 	"github.com/spf13/cobra"
 
 	"github.com/ocurity/dracon/cmd/draconctl/components"
 	"github.com/ocurity/dracon/cmd/draconctl/migrations"
 	"github.com/ocurity/dracon/cmd/draconctl/pipelines"
+	draconLogger "github.com/ocurity/dracon/pkg/log"
```
ptzianos, line 12: import aliases should not have uppercase letter according to the Go standard

components/consumers/consumer.go
```diff
@@ -5,47 +5,49 @@ package consumers
 import (
 	"flag"
 	"fmt"
+	"log/slog"
+	"os"
 
-	v1 "github.com/ocurity/dracon/api/proto/v1"
+	draconapiv1 "github.com/ocurity/dracon/api/proto/v1"
+	"github.com/ocurity/dracon/components"
```
ptzianos, line 13: there shouldn't be empty lines between imports that should be in the same group

components/consumers/consumer.go
```diff
@@ -5,47 +5,49 @@ package consumers
...
 func init() {
 	flag.StringVar(&inResults, "in", "", "the directory where dracon producer/enricher outputs are")
 	flag.BoolVar(&Raw, "raw", false, "if the non-enriched results should be used")
+	flag.BoolVar(&Debug, "debug", false, "turn on debug logging")
 }
 
 // ParseFlags will parse the input flags for the consumer and perform simple validation.
 func ParseFlags() error {
 	flag.Parse()
+	draconLogger.SetDefault(slog.LevelInfo, os.Getenv(components.EnvDraconScanID), true)
```
ptzianos, line 35: the only thing that is changing here is the logging level. it should be assigned to a variable based on the flag and then the `SetDefault` should be invoked once

components/producers/producer.go
```diff
@@ -11,13 +11,16 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"log/slog"
```
ptzianos, line 14: we added `slog` so not replace `log` completely in the file

components/consumers/consumer.go
```diff
@@ -5,47 +5,49 @@ package consumers
...
-	// EnvDraconScanTags the tags of the dracon scan.
-	EnvDraconScanTags = "DRACON_SCAN_TAGS"
-)
-
 var (
 	inResults string
 	// Raw represents if the non-enriched results should be used.
 	Raw bool
+	// Debug flag initializes the logger with a debug level
+	Debug bool
```
ptzianos, line 23: why is Raw and Debug public over here? who needs access to the values of these variables? `Raw` also doesn't seem to be used at all

components/consumers/consumer_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	v1 "github.com/ocurity/dracon/api/proto/v1"
+	"github.com/ocurity/dracon/components"
```
ptzianos, line 10: these imports should be after the third party imports

components/producers/producer.go
```diff
@@ -28,26 +31,26 @@ var (
...
 	// Append flag will append to the outfile instead of overwriting, useful when there's multiple inresults.
 	Append bool
+	// Debug flag initializes the logger with a debug level
+	Debug bool
 )
 
 const (
 	sourceDir = "/workspace/output"
-
-	// EnvDraconStartTime Start Time of Dracon Scan in RFC3339.
```
ptzianos, line 36: why where these consts moved? they don't seem to have any influence related to the logging stuff

northdpole: because  they were duplicates, not necessarily part of THIS PR but bear with me, it's tiny lets merge it

components/producers/producer.go
```diff
@@ -113,28 +117,28 @@ func ParseMultiJSONMessages(in []byte) ([]interface{}, error) {
...
 	}
-	scanUUUID := strings.TrimSpace(os.Getenv(EnvDraconScanID))
-	scanTagsStr := strings.TrimSpace(os.Getenv(EnvDraconScanTags))
+	scanUUUID := strings.TrimSpace(os.Getenv(components.EnvDraconScanID))
+	scanTagsStr := strings.TrimSpace(os.Getenv(components.EnvDraconScanTags))
 	scanTags := map[string]string{}
 	err := json.Unmarshal([]byte(scanTagsStr), &scanTags)
 	if err != nil {
-		log.Printf("scan with uuid %s does not have any tags, err:%s", scanUUUID, err)
+		slog.Info(fmt.Sprintf("scan with uuid %s does not have any tags, err:%s", scanUUUID, err))
```
ptzianos, line 141: This doesn't look like a correct use of the logger. the default logger setup will add the uuid as a key value pair in the log entry. Also this should be an invocation of `Error` instead of `Info`

pkg/log/dracon-logger.go
```diff
@@ -0,0 +1,19 @@
+package draconLogger
+
+import (
+	"log/slog"
+	"os"
+)
+
+func SetDefault(logLevel slog.Level, scanID string, jsonLogging bool) {
```
ptzianos, line 8: this function has no docstring, the linter eventually will start complaining

components/producers/producer.go
```diff
@@ -153,7 +157,7 @@ func getSource() string {
 
 	dat, err := os.ReadFile(sourceMetaPath)
 	if err != nil {
-		log.Println(err)
+		slog.Error(err.Error())
```
ptzianos, line 160: I don't think you need to pass the `.Error()`, internally the `String()` method will be invoked and the result will be the same

northdpole: i need to pass the error, slog only logs strings :(

ptzianos (approved):

PR #193, opened 2024-06-03
make migrations job able to be deleted and recreated many times so th…

northdpole:
…at we can deploy multiple times


dlicheva (approved):

ptzianos (approved):

ptzianos:
there should be a ticket related to this with a description of what is the expected outcome. also it's better to have a short commit message and a commit description

northdpole:
> there should be a ticket related to this with a description of what is the expected outcome. also it's better to have a short commit message and a commit description

thanks for the review, given the tiny change i thought the bytes and work required for all of those were more than the code itself

PR #206, opened 2024-06-06
[OCU-118] 🐛 Fix Semgrep and Bandit producers not recording CWE

flowirtz:
Both the Semgrep and Bandit producers were not correctly propagating the CWE from their JSON inputs. This is now fixed.

### DIFF
--- a/components/producers/python-bandit/main.go
+++ b/components/producers/python-bandit/main.go
@@ -59,6 +59,7 @@ func parseResult(r *BanditResult) (*v1.Issue, error) {
 		Cvss:        0.0,
 		Confidence:  v1.Confidence(v1.Confidence_value[fmt.Sprintf("CONFIDENCE_%s", r.IssueConfidence)]),
 		Description: fmt.Sprintf("%s\ncode:%s", r.IssueText, r.Code),
+		Cwe:         []int32{r.IssueCWE.ID},
 	}
 
 	// Extract the code snippet, if possible
@@ -92,6 +93,9 @@ type BanditResult struct {
 	MoreInfo        string   `json:"more_info"`
 	TestID          string   `json:"test_id"`
 	TestName        string   `json:"test_name"`
+	IssueCWE        struct {
+		ID int32 `json:"id"`
+	} `json:"issue_cwe"`
 }
 
 // // BanditMetric represents a Bandit Metric
--- a/components/producers/semgrep/main.go
+++ b/components/producers/semgrep/main.go
@@ -55,6 +55,7 @@ func parseIssues(out types.SemgrepResults) ([]*v1.Issue, error) {
 		}
 
 		sev := severityMap[r.Extra.Severity]
+
 		iss := &v1.Issue{
 			Target:      producers.GetFileTarget(r.Path, r.Start.Line, r.End.Line),
 			Type:        r.Extra.Message,
@@ -63,6 +64,7 @@ func parseIssues(out types.SemgrepResults) ([]*v1.Issue, error) {
 			Cvss:        0.0,
 			Confidence:  v1.Confidence_CONFIDENCE_MEDIUM,
 			Description: fmt.Sprintf("%s\n extra lines: %s", r.Extra.Message, r.Extra.Lines),
+			Cwe:         r.Extra.Metadata.CWE,
 		}
 
 		// Extract the code snippet, if possible
--- a/components/producers/semgrep/main_test.go
+++ b/components/producers/semgrep/main_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 
 	v1 "github.com/ocurity/dracon/api/proto/v1"
-	"github.com/ocurity/dracon/components/producers"
 	types "github.com/ocurity/dracon/components/producers/semgrep/types"
 	"github.com/ocurity/dracon/pkg/testutil"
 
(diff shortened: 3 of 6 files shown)


components/producers/semgrep/main.go
```diff
@@ -72,3 +75,41 @@ func parseIssues(out types.SemgrepResults) ([]*v1.Issue, error) {
...
+	case []interface{}:
+		for _, s := range v {
+			cweAsInt := convertCWEStringToInt(s.(string))
+			cweInts = append(cweInts, int32(cweAsInt))
+		}
+	case string:
+		cweAsInt := convertCWEStringToInt(v)
+		cweInts = append(cweInts, int32(cweAsInt))
+	default:
+		log.Fatalf("invalid cwe type: %T", cwe)
```
ptzianos, line 93: we should not use log.Fatalf anywhere outside of a main, we should return an error instead and let the caller decide if they want to stop executing or keep going. there are a bunch of `log.Fatal` used also further down that should be fixed

components/producers/python-bandit/main.go
```diff
@@ -87,6 +88,9 @@ type BanditResult struct {
 	MoreInfo        string   `json:"more_info"`
 	TestID          string   `json:"test_id"`
 	TestName        string   `json:"test_name"`
+	IssueCWE        struct {
+		ID int32 `json:"id"`
```
ptzianos, line 92: why is this specifically an int32? the default int Golang type is already 32 bits long, using an int32 just makes everything more verbose when processing it since all the default Golang libraries return `int`

components/producers/semgrep/types/semgrep-issue.go
```diff
@@ -33,4 +33,6 @@ type SemgrepResults struct {
 type Metavars struct{}
 
 // Metadata currently is empty, however, could represent semgrep issue metadata going forward.
-type Metadata struct{}
+type Metadata struct {
+	CWE interface{} `json:"cwe"` // Can be a []string but also just a string
```
ptzianos, line 37: why is this handled in a separate function and we don't override instead the JSON marshaller so that we can have a CWE []int instead of having an `interface{}`

components/producers/python-bandit/main.go
flowirtz: The `Issue.Cwe` in `issue.pb.go` is `[]int32` - do u want me to change that?

components/producers/semgrep/types/semgrep-issue.go
flowirtz: No idea how that would work or what that would look like, tbh - can u give me any pointers?

components/producers/semgrep/main.go
flowirtz: fixed! thanks

components/producers/semgrep/types/semgrep-issue.go
ptzianos: https://pkg.go.dev/encoding/json#Marshaler

flowirtz: so you mean:

1. change the `Metadata.CWE: int[]`
2. in the `main`, do a
```diff
- if err := json.Unmarshal(inFile, &results); err != nil {
+ if err := ourCustomJSONUnmarshalThatHandlesTheCWE(...`
```

where
```golang
func ourCustomJSONUnmarshalThatHandlesTheCWE(...) {
  json.Unmarshal(...)
  // also handle the interface{} -> int[] conversion
}
```

is that what you mean?

ptzianos: Yes we should use properly types fields unless there is a very good reason not to. I have rarely seen a good reason to use the `interface{}`.
No, that's not what the docs of the json library say. There are plenty of examples of how you implement custom JSON marshaling/unmarshaling in Golang: https://gobyexample.com/json.

components/producers/python-bandit/main.go
ptzianos: I didn't notice that we are using `int32`, I don't see a reason to honestly, but it's going to be a hassle to change the protos, so let's stick with it for the time being and we revisit it in the future.

components/producers/semgrep/types/semgrep-issue.go
```diff
@@ -32,5 +40,76 @@ type SemgrepResults struct {
...
 
-// Metadata currently is empty, however, could represent semgrep issue metadata going forward.
-type Metadata struct{}
+// Metadata contains semgrep issue metadata
+type Metadata struct {
+	CWE []int32
+}
+
+func (t *Metadata) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" || string(data) == `""` {
```
ptzianos, line 49: you should simply check if `data == nil`

components/producers/semgrep/types/semgrep-issue.go
```diff
@@ -32,5 +40,76 @@ type SemgrepResults struct {
...
+		return nil
+	}
+
+	var realMetadata struct {
```
ptzianos, line 53: I don't understand why this is so complex. As far as I see you are checking to see if you have one `CWE-[0-9]+` or multiple elements in a list. You could just check if the data you get converted into a string starts with a `CWE` otherwise try to unmarshal the bytes into a slice of strings and then convert them into `int32`s

Also there is no test for the correct JSON unmarshaling of the Metadata object

components/producers/semgrep/main_test.go
```diff
@@ -25,7 +25,11 @@ const exampleOutput = `
 			"extra": {
 				"message": "Use of this type presents a security risk: the encapsulated content should come from a trusted source, \nas it will be included verbatim in the template output.\nhttps://blogtitle.github.io/go-safe-html/\n", 
 				"metavars": {},
-				"metadata": {}, 
+				"metadata": {
+					"cwe": [
+						"CWE-89: Improper Neutralization of Special Elements used in an SQL Command ('SQL Injection')"
```
ptzianos, line 30: can we add a second entry over here to ensure that the code works as expected in the case of a list with multiple entries

ptzianos (approved):

PR #213, opened 2024-06-11
[OCU-122] ✨ Add new `GetPURLTarget` method to base-producer

flowirtz:
This PR adds support for pURL `Issue.target`s to the base-producer.

We're introducing two new methods:
- `GetPURLTarget` to construct a valid pURL from it's parts (e.g. package name, version, etc.)
- `EnsureValidPURLTarget` to ensure a given pURL is valid, e.g. if it's supplied by a scanner already

Additionally, we are switching over three SCA-type producers to start recording their targets as valid pURLs. The producers being switched are `docker-trivy`, `dependency-check` and `python-pip-safety`. There are two more SCA producers in this repository which we will switch over shortly [via OCU-125](https://linear.app/smithy/issue/OCU-125/switch-over-remaining-sca-producers-to-produce-purl-targets).

Lastly, we are also applying a new metadata label in the corresponding `task.yaml` files to denote a producer being a SCA producer via `v1.dracon.ocurity.com/test-type: sca`.


flowirtz:
FYI Linter fails, as discussed with @ptzianos - we're disabling it in this PR anyways. Works locally.

components/producers/dependency-check/examples/README.md
```diff
@@ -0,0 +1,12 @@
+# Examples
+
+## requirements.txt
+
+```bash
+brew update && brew install dependency-check
```
northdpole, line 6: is this mac-specific?

components/producers/dependency-check/examples/requirements.txt
```diff
@@ -0,0 +1,20 @@
+Django==1.11.29
```
northdpole, line 1: what's this file used for?

northdpole, line 1: just groked that this is used as example vulnerable input can you please add a 
`\# comment at the top saying that this is example input data to dependency-track`
? also would be good to add a couple known vulnerable pakcages so we know what to expect

components/producers/producer.go
```diff
@@ -160,3 +162,21 @@ func getSource() string {
...
+//
+// Example: GetPURLTarget("deb", "debian", "curl", "7.68.0", nil, "")
+func GetPURLTarget(purlType string, namespace string, name string, version string, qualifiers packageurl.Qualifiers, subpath string) string {
+	return packageurl.NewPackageURL(purlType, namespace, name, version, qualifiers, subpath).ToString()
+}
+
+// EnsureValidPURLTarget takes a purl target string from an untrusted source, e.g. a tool output, and
+// ensures it is a valid purl target according to the packageurl-go library.
+func EnsureValidPURLTarget(purlTarget string) (string, error) {
+	instance, err := packageurl.FromString(purlTarget)
```
northdpole, line 177:
```suggestion
	_, err := packageurl.FromString(purlTarget)
```

components/producers/producer.go
```diff
@@ -160,3 +162,21 @@ func getSource() string {
...
+	if err != nil {
+		return "", err
+	}
+	return instance.ToString(), nil
```
northdpole, line 181:
```suggestion
	return purlTarget, nil
```

components/producers/producer_test.go
```diff
@@ -121,3 +123,29 @@ func TestParseMultiJSONMessages(t *testing.T) {
 	}
 	assert.Equal(t, want[0].Foo, "bar")
 }
+
+func TestGetPURLTarget(t *testing.T) {
+	target := GetPURLTarget("deb", "debian", "curl", "7.68.0", nil, "")
+	assert.Equal(t, "pkg:deb/debian/curl@7.68.0", target)
```
northdpole, line 129:
```suggestion
	require.Equal(t, "pkg:deb/debian/curl@7.68.0", target)
```

components/producers/producer_test.go
```diff
@@ -121,3 +123,29 @@ func TestParseMultiJSONMessages(t *testing.T) {
...
+
+	target = GetPURLTarget("bitbucket", "birkenfeld", "pygments-main", "244fd47e07d1014f0aed9c", nil, "")
+	assert.Equal(t, "pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c", target)
```
northdpole, line 132:
```suggestion
	require.Equal(t, "pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c", target)
```

assertions package introduced `require` which allows for immediate exit and less noise, can we please use that?

components/producers/producer_test.go
```diff
@@ -121,3 +123,29 @@ func TestParseMultiJSONMessages(t *testing.T) {
...
+
+	target = GetPURLTarget("docker", "customer", "dockerimage", "sha256:244fd47e07d1004f0aed9c", packageurl.Qualifiers{
+		{Key: "repository_url", Value: "gcr.io"},
+	}, "")
+	assert.Equal(t, "pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io", target)
```
northdpole, line 137:
```suggestion
	require.Equal(t, "pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io", target)
```

components/producers/producer_test.go
```diff
@@ -121,3 +123,29 @@ func TestParseMultiJSONMessages(t *testing.T) {
...
+
+	target = GetPURLTarget("npm", "", "foobar", "12.3.1", nil, "")
+	assert.Equal(t, "pkg:npm/foobar@12.3.1", target)
```
northdpole, line 140:
```suggestion
	require.Equal(t, "pkg:npm/foobar@12.3.1", target)
```

components/producers/producer_test.go
```diff
@@ -121,3 +123,29 @@ func TestParseMultiJSONMessages(t *testing.T) {
...
+
+	target = GetPURLTarget("pypi", "", "django", "1.11.1", nil, "")
+	assert.Equal(t, "pkg:pypi/django@1.11.1", target)
```
northdpole, line 143:
```suggestion
	require.Equal(t, "pkg:pypi/django@1.11.1", target)
```

components/producers/producer_test.go
```diff
@@ -121,3 +123,29 @@ func TestParseMultiJSONMessages(t *testing.T) {
...
+
+	target = GetPURLTarget("deb", "debian", "curl", "7.50.3-1", packageurl.Qualifiers{
+		{Key: "arch", Value: "i386"},
+		{Key: "distro", Value: "jessie"},
+	}, "")
+
+	assert.Equal(t, "pkg:deb/debian/curl@7.50.3-1?arch=i386&distro=jessie", target)
```
northdpole, line 150:
```suggestion
	require.Equal(t, "pkg:deb/debian/curl@7.50.3-1?arch=i386&distro=jessie", target)
```

components/producers/producer_test.go
```diff
@@ -121,3 +123,29 @@ func TestParseMultiJSONMessages(t *testing.T) {
...
+}
```
northdpole, line 151: can you also please add an test for the error case?
also, you don't need to test the third party library so much, i hope they have their own test cases :p 
you can do so with `require.Error()`

components/producers/python-pip-safety/examples/requirements.txt
```diff
@@ -0,0 +1,20 @@
+Django==1.11.29
```
northdpole, line 1: same applies for this requirements file, please add a comment stating what it is for and mark the offending deps

components/producers/python-pip-safety/main.go
```diff
@@ -12,13 +12,13 @@ import (
 
 func parseIssues(out []types.Vulnerability) []*v1.Issue {
 	issues := []*v1.Issue{}
+
 	for _, r := range out {
 		issues = append(issues, &v1.Issue{
-			Target:     fmt.Sprintf("%s:%s", r.PackageName, r.AnalyzedVersion),
+			Target:     producers.GetPURLTarget("pypi", "", r.PackageName, r.AnalyzedVersion, nil, ""),
```
northdpole, line 18: hint, instead of the string "pypi", you can use [the relevant type from package-url go](https://github.com/package-url/packageurl-go/blob/master/packageurl.go#L54)

components/producers/test_utils.go
```diff
@@ -0,0 +1,70 @@
+package producers
```
northdpole, line 1: this is a good idea, please open an RFC PR or a ticket or "something" where you document a testing approach.
we can then all pitch in and ensure that current and future components have the correct structure they need to in order to be testable

northdpole (dismissed):

components/producers/dependency-check/examples/README.md
flowirtz: no, actually you can use `brew` on linux too! But yeah, agreed - I'll tweak the wording

components/producers/producer.go
flowirtz: mmh, no, I'd rather use the string where we validated it's correct, instead of the input. The implementation of `FromString` does a couple modifications on the input, so it's possible that the input changes

components/producers/producer_test.go
flowirtz: actually there is no error, because of the types - it just passes everything down via the constructor, and then strings things together

northdpole (approved):

PR #218, opened 2024-06-12
[OCU-126] 🏷️ Extend `LaunchToolResponse` proto by new `scan_target` field

flowirtz:
This PR extends the `LaunchToolResponse` with a new field `scan_target`.

Moving forward, we will use this new field to record which target a scan was directed at. More information can be found in [OCU-126](https://linear.app/smithy/issue/OCU-126/extend-launchtoolresponse-to-have-target-field).

Additionally, we are adding a missing `buf.gen.yaml` that is used to compile the protobufs to go. We also updated the docs for future reference.


docs/building-protos.md
```diff
@@ -0,0 +1,16 @@
+# Building Protobufs
+
+We use [buf.build](https://buf.build/) to simplify compiling the various `.proto` files.
+You can install and compile the protos via:
+
+```bash
+make install-lint-tools
+buf generate
```
northdpole, line 8: maybe a nit:
can you please add a `make protos` target?

ptzianos:
other than the nit Spyros suggested I think it looks good

ptzianos:
once it's fixed and merged please tag the commit

northdpole (approved):

PR #219, opened 2024-06-12
Move draconctl migrations apply <path> to environment variable

### LINKED ISSUE #220: Allow draconctl to accept migrations path from environment variable
since draconctl migrations is mostly supposed to be used in a k8s job it makes the most sense for hte path of the migrations to be supplied via environment variable

northdpole:
closes #220 

can be tested with

`DRACONCTL_MIGRATIONS_PATH=/tmp/migrations go run cmd/draconctl/main.go migrations apply`

notice how the error is not "migrations not found" anymore


tests/migrations/compose.yaml
```diff
@@ -31,13 +31,13 @@ services:
     - |
       apt-get update && apt-get install -y jq &&
       echo "------------------------------ Starting tests ------------------------------" &&
-      /bin/draconctl migrations inspect --url postgres://postgres:password@postgres:5432?sslmode=disable --json /etc/dracon/migrations/enrichment | tee | jq -eM '. | select(.Version == 0)' &&
+      DRACONCTL_MIGRATIONS_PATH='/etc/dracon/migrations/enrichment' /bin/draconctl migrations inspect --url postgres://postgres:password@postgres:5432?sslmode=disable --json /etc/dracon/migrations/enrichment | tee | jq -eM '. | select(.Version == 0)' &&
```
ptzianos, line 34: you can export the variable only one time at the top of the script here

docs/getting-started.md
```diff
@@ -255,7 +255,7 @@ bin/cmd/draconctl migrations apply \
   --as-k8s-job \
   --image "${CONTAINER_REPO}/draconctl:${CUSTOM_DRACON_VERSION}" \
   --url "postgresql://dracon:dracon@dracon-enrichment-db.dracon.svc.cluster.local?sslmode=disable" \
-  /etc/dracon/migrations/enrichment
+  --migrations-path /etc/dracon/migrations/enrichment
```
ptzianos, line 258: this is going to be run inside a migration job, why not have it as an env variable inside the container?

cmd/draconctl/migrations/migrations.go
```diff
@@ -72,6 +73,8 @@ func init() {
 	migrationsCmd.PersistentFlags().StringVarP(&migrationsAsK8sJobConfig.image, "image", "i", "", "Image to use containing draconctl binary to run command")
 	migrationsCmd.PersistentFlags().StringVar(&migrationsAsK8sJobConfig.ssa, "ssa-name", "draconctl", "Name to use for server-side apply")
 	// migrationsCmd.Flags().Bool("provide-password", false, "Provide the password via a console")
+	migrationsCmd.PersistentFlags().StringVar(&migrationsCmdConfig.migratiosnPath, "migrations-path", "", "absolute path of where to find the migrations")
```
ptzianos, line 76: why absolute? why won't a relative path work?

tests/migrations/compose.yaml
```diff
@@ -29,12 +29,13 @@ services:
...
     - -c
     - |
+      export DRACONCTL_MIGRATIONS_PATH='/etc/dracon/migrations/enrichment'
       apt-get update && apt-get install -y jq &&
       echo "------------------------------ Starting tests ------------------------------" &&
       /bin/draconctl migrations inspect --url postgres://postgres:password@postgres:5432?sslmode=disable --json /etc/dracon/migrations/enrichment | tee | jq -eM '. | select(.Version == 0)' &&
       echo "------------------------------ Applying migrations ------------------------------" &&
-      /bin/draconctl migrations apply --url postgres://postgres:password@postgres:5432?sslmode=disable /etc/dracon/migrations/enrichment &&
-      /bin/draconctl migrations inspect --url postgres://postgres:password@postgres:5432?sslmode=disable --json /etc/dracon/migrations/enrichment | tee | jq -eM '. | select(.Version > 0)' &&
+      /bin/draconctl migrations apply --url postgres://postgres:password@postgres:5432?sslmode=disable --migrations-path /etc/dracon/migrations/enrichment &&
```
ptzianos, line 37: I don't think the flag is needed any more, since the command should pick up the correct path from the environment variable?

docs/getting-started.md
```diff
@@ -250,12 +250,13 @@ kubectl apply -n dracon -f deploy/dracon/serviceaccount.yaml
 kubectl apply -n dracon -f deploy/dracon/role.yaml
 kubectl apply -n dracon -f deploy/dracon/rolebinding.yaml
 make cmd/draconctl/bin
+
+export DRACONCTL_MIGRATIONS_PATH='/etc/dracon/migrations/enrichment'
```
ptzianos, line 254: is this value going to be passed to the container? I don't see any way in the code to propagate it. Also I don't think that it's needed when the dracon migrations container is used. maybe we can have it in the Dockerfile of the migrations container, since we are copying the migrations folder in there

ptzianos (approved):

PR #230, opened 2024-07-05
[OCU-113] 🐛 Fix `producer/semgrep` not supporting registry

flowirtz:
This PR adds a new parameter `producer-semgrep-config-value` to the `producer/semgrep`. By default, it's set to `auto`.

Adding this new parameter makes the SemGrep producer immediately useful, without needing to define custom rules in YAML.

The binary also supports multiple `--config` parameters, so the two parameters don't interfere with each other.


northdpole (approved):
please add an example, i have no idea how to run this :)

northdpole (approved):

PR #231, opened 2024-07-08
[OCU-125] Add pURL targets for remaining SCA producers

flowirtz:
This PR switches over the two last remaining SCA producers that didn't yet produce pURLs as `Target`s.

For both the `producers/golang-nancy` and `producers/typescript-yarn-audit`, this PR is:
- switching them over to record pURLs as the `Issue.Target`
- adding support for them to record CWEs
- adding example data and documentation
- adding an e2e test
- adding a `SCA` label as the `test-type` in the `task.yaml`

Additionally, this PR removes the `producers/typescript-npm-audit` [as discussed here](https://ocurity.slack.com/archives/C06D2AAE51R/p1720518498837449).


components/producers/producer.go
```diff
@@ -62,24 +61,6 @@ func ParseFlags() error {
...
-	for scanner.Scan() {
-		result = append(result, scanner.Bytes())
-	}
-
-	return result, err
-}
-
 // ReadInFile returns the contents of the file given by InResults.
 // TODO: replace with os.ReadFile
 func ReadInFile() ([]byte, error) {
```
ptzianos, line 66: I added a TODO above this line so that we can tackle this eventually.

The `ReadInFile` function is a (bad) re-implementation of the [os.ReadFile](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/os/file.go;l=693). Since we are refactoring this code we should get rid of this function too.

components/producers/golang-nancy/main_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	v1 "github.com/ocurity/dracon/api/proto/v1"
```
ptzianos, line 8: these imports are our own, they should be the last block, we should have first the standard library ones, then a block with the third parties, then our own

components/producers/golang-nancy/main.go
```diff
@@ -35,6 +44,17 @@ func main() {
 	}
 }
 
+func getCWEFromTitle(title string) string {
+	// Input would look like the below
+	// [CVE-2023-26125] CWE-20: Improper Input Validation
+	re := regexp.MustCompile(`CWE-\d+`)
```
ptzianos, line 50: we should have this as a package variable so that if there is an issue with it, it will explode as soon as the binary executes instead of whenever this piece of code is invoked.

components/producers/golang-nancy/main_test.go
```diff
@@ -145,3 +146,16 @@ var exampleOutput = `{
 	]
   }
 `
+
+func TestGetCWEFromTitle(t *testing.T) {
+	cwe := getCWEFromTitle("[CVE-2023-26125] CWE-20: Improper Input Validation")
+	assert.Equal(t, "CWE-20", cwe)
```
ptzianos, line 152: we should `require` instead of `assert`. If there is an error `assert` wont' stop the test, also this should be implemented as a table driven test

components/producers/producer.go
flowirtz: fair point! missed that, thanks

northdpole (approved):

PR #236, opened 2024-07-11
[OCU-104] Ensure Enrichers Always Produce an Output

flowirtz:
Via OCU-104 we discovered that the `enricher-aggregator` component crashes if no issues were enriched. This is a bug. It shouldn't crash, it should simply produce an aggregated output with no findings.

In this PR, we ensure all existing enricher components always produce a result. If there were no issues supplied to the respective enrichers we simply store an empty list result.

 We also add tests for this behaviour to all enrichers.

### DIFF
--- a/components/enrichers/enricher.go
+++ b/components/enrichers/enricher.go
@@ -58,18 +58,25 @@ func LoadData() ([]*draconV1.LaunchToolResponse, error) {
 
 // WriteData will write the enriched results to the write path.
 func WriteData(enrichedLaunchToolResponse *draconV1.EnrichedLaunchToolResponse, enricherName string) error {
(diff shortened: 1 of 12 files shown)


flowirtz:
Blocked by #192

components/enrichers/enricher.go
```diff
@@ -58,18 +58,25 @@ func LoadData() ([]*draconV1.LaunchToolResponse, error) {
 
 // WriteData will write the enriched results to the write path.
 func WriteData(enrichedLaunchToolResponse *draconV1.EnrichedLaunchToolResponse, enricherName string) error {
-	if enrichedLaunchToolResponse == nil || len(enrichedLaunchToolResponse.Issues) == 0 {
```
northdpole, line 61: why is this removed? wouldn't a valid TR with no issues still throw this error?

northdpole (approved):

PR #238, opened 2024-07-12
[OCU-113] make draconctl able to log JSON

### LINKED ISSUE #191: Provide draconctl with the ability to log in json format
this ticket is about adding a flag to the rootCMD of` --log-format {text,json}` which when enabled sets the default logger to either plaintext (default) or json<br>Useful when running draconctl in a cluster

northdpole:
closes #191 
adds ` --log-format {text,json} `  flag to rootCMD which when enabled sets the default logger to either plaintext (default) or json
Useful when running draconctl in a cluster

### DIFF
--- a/cmd/draconctl/main.go
+++ b/cmd/draconctl/main.go
@@ -2,11 +2,12 @@ package main
 
 import (
 	"fmt"
-	"log"
 	"log/slog"
 	"os"
+	"slices"
 	"strings"
 
+	"github.com/go-errors/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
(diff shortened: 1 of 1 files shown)


cmd/draconctl/main.go
```diff
@@ -58,6 +66,12 @@ func initializeConfig(cmd *cobra.Command) error {
 	// like --favorite-color which we fix in the bindFlags function
 	v.AutomaticEnv()
 
+	if !slices.Contains(supportedLogFormats, strings.ToLower(loggingFormat)) {
+		return errors.Errorf("unsupported log format %s, supported formats are %s", loggingFormat, supportedLogFormats)
+	} else if strings.ToLower(loggingFormat) == "json" {
+		slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{})))
```
ptzianos, line 72: since we are switching to `slog` we should also replace the `log` invocation a few lines below this

cmd/draconctl/main.go
```diff
@@ -39,7 +45,9 @@ func main() {
 		os.Exit(1)
 	}
 }
-
+func init() {
+	rootCmd.PersistentFlags().StringVar(&loggingFormat, "log-format", "text", "The log format to use (supports 'text' for plain text and 'json' for JSON)")
+}
```
ptzianos, line 50: there should be an empty line between the functions

PR #269, opened 2024-07-18
[OCU-155] Fix example workflows are failing

flowirtz:
This PR fixes three existing example workflows, so they run successfully and produce valid results.

image

We now have three example pipelines (golang, typescript, python) that all produce useful results. To achieve this, several things were changed.

## Config Changes
- 🔧 Change default ES and MongoDB URLs: They were referencing nonexistent instances
- 🔧 Replace enricher-deduplication with enricher-codeowners: De-duplication is broken atm, Spyros is working on a fix
- 🔧 Change example golang pipeline: Referenced a few params that aren't needed
- 🔧 Change example python pipeline: Also referenced a few params that aren't needed

## Bug Fixes
- 🐛 Fix incorrect parameter name in TS example pipeline: Was referencing a nonexistent parameter
- 🐛 Fix producer/yarn-audit not handling empty lines: The input file is a very weird JSON, sometimes it has empty lines, we crashed previously
- 🐛 Fix erroneous image reference in eslint producer: The image didn't exist
- 🐛 Fix producer/eslint component not starting
- 🐛 Fix producer/semgrep fails to run
- 🐛 Fix typescript example workflow
- 🐛 Fix enrichers not handling multiple tools

## Misc
- 🔨 Add utility script to bump local components
- ♻️ Refactor image pinning test

### DIFF
--- a/components/enrichers/codeowners/main.go
+++ b/components/enrichers/codeowners/main.go
@@ -81,10 +81,13 @@ func run() error {
 			enrichedIssues = append(enrichedIssues, eI)
 		}
 
-		return enrichers.WriteData(&apiv1.EnrichedLaunchToolResponse{
+		err := enrichers.WriteData(&apiv1.EnrichedLaunchToolResponse{
 			OriginalResults: r,
 			Issues:          enrichedIssues,
 		}, "codeowners")
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ GO_TEST_PACKAGES=$(shell go list ./... | grep -v /vendor/)
 CONTAINER_REPO=ghcr.io/ocurity/dracon
 SOURCE_CODE_REPO=https://github.com/ocurity/dracon
 DRACON_VERSION=$(shell echo $(latest_tag)$$([ $(commits_since_latest_tag) -eq 0 ] || echo "-$$(git log -n 1 --pretty='format:%h')" )$$([ -z "$$(git status --porcelain=v1 2>/dev/null)" ] || echo "-dirty" ))
+DRACON_OSS_COMPONENTS_NAME=dracon-oss-components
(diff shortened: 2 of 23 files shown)


scripts/update_local_oss_components.sh
```diff
@@ -0,0 +1,23 @@
...
+# instance. It will build all components based on the latest local commit and 
+# update the local dracon instance with the new components.
+
+echo "Updating open-source components in local dracon instance..."
+export CUSTOM_DRACON_VERSION=$(make print-DRACON_VERSION)
+echo "DRACON_VERSION=$CUSTOM_DRACON_VERSION"
+export CUSTOM_HELM_COMPONENT_PACKAGE_NAME="dracon-oss-components"
+
+make publish-component-containers CONTAINER_REPO=localhost:5000/ocurity/dracon
+bin/cmd/draconctl components package   --version ${CUSTOM_DRACON_VERSION}   \
```
ptzianos, line 13: why not have this as a dev make target?

components/enrichers/codeowners/main.go
```diff
@@ -81,10 +81,13 @@ func run() error {
 			enrichedIssues = append(enrichedIssues, eI)
 		}
 
-		return enrichers.WriteData(&apiv1.EnrichedLaunchToolResponse{
+		err := enrichers.WriteData(&apiv1.EnrichedLaunchToolResponse{
```
northdpole, line 84: how is this different from before? it still returns either an err or nil

northdpole (approved):

components/enrichers/codeowners/main.go
flowirtz: ha, I'm glad you asked! It's true, it still either returns `err` or `nil`.

But: It's inside a loop. The `return` always exits that loop _on the first iteration_.
The enricher never got to the second or third `LaunchToolResponse` that it was meant to parse.

PR #276, opened 2024-07-29
fix deduplication db default postgresql by setting sane defaults and …

northdpole:
…allowing overriding


components/enrichers/deduplication/task.yaml
```diff
@@ -20,4 +20,4 @@ spec:
     - name: ENRICHER_WRITE_PATH
       value: "$(workspaces.output.path)/.dracon/enrichers/deduplication"
     - name: ENRICHER_DB_CONNECTION
-      value: "postgresql://dracon:dracon@dracon-enrichment-db.$(context.taskRun.namespace).svc?sslmode=disable"
+      value: postgresql://{{default "dracon" .Values.postgresql.auth.username }}:{{default "dracon" .Values.postgresql.auth.password }}@{{ default "dracon-enrichment-db.dracon.svc" .Values.postgresql.host }}/{{default "dracon" .Values.postgresql.auth.database }}?{{default "sslmode=disable" .Values.postgresql.auth.querystringargs}}
```
flowirtz, line 23: Just had this discussion with @ptzianos over [here](https://github.com/ocurity/dracon/pull/269#discussion_r1691035515): The Conclusion was that we want the default connection URLs not to be namespaced explicitly, but rather assume they're in the same namespace. This way people can use namespaces other than `dracon` and it's easier for us to switch the default ns in the future.

I.e., instead of `dracon-enrichment-db.dracon.svc` just do `<service-name>` so `dracon-enrichment-db`
```suggestion
      value: postgresql://{{default "dracon" .Values.postgresql.auth.username }}:{{default "dracon" .Values.postgresql.auth.password }}@{{ default "dracon-enrichment-db" .Values.postgresql.host }}/{{default "dracon" .Values.postgresql.auth.database }}?{{default "sslmode=disable" .Values.postgresql.auth.querystringargs}}
```

flowirtz (approved):
one comment

flowirtz (approved):

flowirtz (approved):

PR #279, opened 2024-07-30
Improve Go Test Runner

flowirtz:
This PR improves the golang test runner, both in CI but also locally. In particular, it does these three things to achieve that:

### 👷 Replace `go test` with `gotestsum`
Previously our test output was only in JSON - easy to read for a machine, harder to read for a human.
The `gotestsum` package is a drop-in replacement for `go test` and [very widely used](https://github.com/gotestyourself/gotestsum?tab=readme-ov-file#who-uses-gotestsum) (e.g. by kubernetes).

Three main improvements it brings are:
1. Change the [test output format](https://github.com/gotestyourself/gotestsum?tab=readme-ov-file#output-format), from compact to verbose with color highlighting.
2. Print a [summary](https://github.com/gotestyourself/gotestsum?tab=readme-ov-file#summary) of the test run after running all the tests.
3. write a JUnit XML file for integration with CI systems (used for test summary below)

example test output

### 👷 Add test summary for Go tests
GitHub previously introduced job summaries, which among other things, provides richer information like shown below rather than just a simple `pass/fail`. We are now exporting that information to GHA as well.
img

### ⬆️ Bump `actions/setup-go@v4` to `v5`
Previously we kept getting this deprecation warning, which is [further explained here](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default). By bumping the action to `v5` that's resolved.
image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,10 @@ name: Test
 on:
   push:
     branches:
-    - main
+      - main
   pull_request:
     branches:
-    - main
+      - main
 
 permissions:
   # Grant the ability to checkout the repository
@@ -27,6 +27,9 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version-file: go.mod
+
+      - name: Install Go Test Tools
+        run: make install-go-test-tools
 
       - name: Run Go tests
         run: make go-tests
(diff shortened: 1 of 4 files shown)


dlicheva (approved):
It looks good! Is it worth adding a note to the README, to tell the user that they need to run `make install-go-test-tools` before testing?

.github/workflows/test.yml
```diff
@@ -43,3 +43,9 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           format: golang
           file: tests/output/cover.out
+
+      - name: Test Summary
+        uses: test-summary/action@v2
+        with:
+          paths: "tests/output/*-tests.xml"
+        if: always()
```
ptzianos, line 51: why is this if needed over here?

flowirtz: it's something [they recommended in their docs](https://github.com/test-summary/action?tab=readme-ov-file#getting-started):
> Note the `if: always()` conditional in this workflow step: you should always use this so that the test summary creation step runs even if the previous steps have failed. This allows your test step to fail -- due to failing tests -- but still produce a test summary.

PR #283, opened 2024-08-01
[OCU-138] Add fileURI support to additional SAST producers

flowirtz:
This PR switches over two more SAST producers to record their targets as fileURIs.


components/producers/producer.go
```diff
@@ -170,8 +170,11 @@ func GetFileTarget(filePath string, startLine int, endLine int) string {
 	if filePath == "" {
 		return ""
 	}
-	url := url.URL{Scheme: "file", Path: filePath}
 
+	// Handle varying prefixes
+	filePath = strings.TrimPrefix(filePath, ".") // Remove leading dot
```
ptzianos, line 175: this doesn't look correct to me. the tests for this expect this path: `file:///path/to/file.txt:1-2`. if the code's current working directory is `/home/app` than I would expect the resulting path to be `file:///home/app/path/to/file.txt:1-2`. Also, there is a function in the standard library that performs this: https://pkg.go.dev/path/filepath#Clean

components/producers/java-findsecbugs/types/xml-types.go
```diff
@@ -0,0 +1,62 @@
...
+	Classname     string   `xml:"classname,attr"`
+	Start         int      `xml:"start,attr"`
+	End           int      `xml:"end,attr"`
+	StartBytecode string   `xml:"startBytecode,attr"`
+	EndBytecode   string   `xml:"endBytecode,attr"`
+	Sourcefile    string   `xml:"sourcefile,attr"`
+	Sourcepath    string   `xml:"sourcepath,attr"`
+	Role          string   `xml:"role,attr"`
+}
+type class struct {
```
ptzianos, line 33: there should be an empty line between structs for readability

components/producers/java-findsecbugs/types/xml-types.go
```diff
@@ -0,0 +1,62 @@
...
+	Classname  string       `xml:"classname,attr"`
+	Role       string       `xml:"role,attr"`
+	SourceLine []SourceLine `xml:"SourceLine"`
+}
+type field struct {
+	XMLName    xml.Name     `xml:"Field"`
+	Classname  string       `xml:"classname,attr"`
+	SourceLine []SourceLine `xml:"SourceLine"`
+}
+type BugInstance struct {
```
ptzianos, line 44: exported structs should have a comment

andream16:
I don't think this is relevant right now - will close.

PR #286, opened 2024-08-06
make oss migrations run as part of dev-dracon

northdpole:
This PR introduces two new ways of installing dracon

` make dev-deploy ` will build everything and deploy it to the local cluster
` make install` will download the latest version of every package and install it to the local cluster

both commands first setup the local cluster in an idempotent way

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+# Developer vars
+# The following variables are used to define the developer environment
+# e.g. what are the test packages, or the latest tag, these are used by make targets that build things
 component_binaries=$(shell find ./components -name main.go | xargs -I'{}' sh -c 'echo $$(dirname {})/bin')
 component_containers=$(shell find ./components -name main.go | xargs -I'{}' sh -c 'echo $$(dirname {})/docker')
 component_containers_publish=$(component_containers:docker=publish)
@@ -5,12 +8,18 @@ protos=$(shell find . -not -path './vendor/*' -name '*.proto')
 go_protos=$(protos:.proto=.pb.go)
 latest_tag=$(shell git tag --list --sort="-version:refname" | head -n 1)
 commits_since_latest_tag=$(shell git log --oneline $(latest_tag)..HEAD | wc -l)
-
 GO_TEST_PACKAGES=$(shell go list ./... | grep -v /vendor/)
+
+# Deployment vars
+#  The following variables are used to define the deployment environment
+# e.g. what are the versions of the components, or the container registry, these are used by make targets that deploy things
 CONTAINER_REPO=ghcr.io/ocurity/dracon
 SOURCE_CODE_REPO=https://github.com/ocurity/dracon
-DRACON_VERSION=$(shell echo $(latest_tag)$$([ $(commits_since_latest_tag) -eq 0 ] || echo "-$$(git log -n 1 --pretty='format:%h')" )$$([ -z "$$(git status --porcelain=v1 2>/dev/null)" ] || echo "-dirty" ))
+DRACON_DEV_VERSION=$(shell echo $(latest_tag)$$([ $(commits_since_latest_tag) -eq 0 ] || echo "-$$(git log -n 1 --pretty='format:%h')" )$$([ -z "$$(git status --porcelain=v1 2>/dev/null)" ] || echo "-dirty" ))
+DRACON_VERSION=$(shell (echo $(CONTAINER_REPO) | grep -q '^ghcr' && echo $(latest_tag)) || echo $(DRACON_DEV_VERSION) )
 DRACON_OSS_COMPONENTS_NAME=dracon-oss-components
+DRACON_OSS_COMPONENTS_PACKAGE_URL=oci://ghcr.io/ocurity/dracon/charts/$(DRACON_OSS_COMPONENTS_NAME)
+
 TEKTON_VERSION=0.44.0
 TEKTON_DASHBOARD_VERSION=0.29.2
 ARANGODB_VERSION=1.2.19
@@ -156,7 +165,8 @@ print-%:
 ########## DEPLOYMENT TARGETS ##########
 ########################################
 .PHONY: deploy-nginx deploy-arangodb-crds deploy-arangodb-operator add-es-helm-repo deploy-elasticoperator \
-		tektoncd-dashboard-helm deploy-tektoncd-dashboard add-bitnami-repo dev-dracon dev-deploy dev-teardown
+		tektoncd-dashboard-helm deploy-tektoncd-dashboard add-bitnami-repo dev-dracon dev-deploy dev-teardown \
+		install install-oss-components deploy-cluster
 
 deploy-nginx:
 	@helm upgrade nginx-ingress https://github.com/kubernetes/ingress-nginx/releases/download/helm-chart-$(NGINX_INGRESS_VERSION)/ingress-nginx-$(NGINX_INGRESS_VERSION).tgz \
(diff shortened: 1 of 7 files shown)


deploy/deduplication-db-migrations/chart/templates/migrations-job.yaml
```diff
@@ -20,7 +20,7 @@ spec:
       restartPolicy: Never
       containers:
       - name: enrichment-db-migrations
-        image: "{{ default "ghcr.io/ocurity/dracon" .Values.image.registry }}/draconctl:{{ default .Chart.AppVersion .Values.image.tag }}"
+        image: "{{ default "ghcr.io/ocurity/dracon" .Values.global.image.registry }}/ocurity/dracon/draconctl:{{ default .Chart.AppVersion .Values.image.tag }}"
```
flowirtz, line 23: this now doesn't work if the default is used, then it'd be `ghcr.io/ocurity/dracon/ocurity/dracon/draconctl...`

deploy/deduplication-db-migrations/chart/values.yaml
```diff
@@ -9,3 +9,5 @@ global:
 
 serviceAccount:
   create: false
+
+enabled: true
```
flowirtz, line 13: why are we enabling it here, rather than via the `values/dev.yaml`?

flowirtz (changes requested):
couple comments

can u pls also delete the `enrichments-db-migrations-0.20-3.tgz` to clarify that this is gone gone?

deploy/deduplication-db-migrations/chart/templates/migrations-job.yaml
flowirtz: u can just move that over from `deploy/dracon/values` and remove it there, it's only needed for the migrations

Makefile
```diff
@@ -219,7 +219,13 @@ dev-dracon: deploy-elasticoperator deploy-arangodb-crds add-bitnami-repo
 		  --values ./deploy/dracon/values/dev.yaml \
 		  --create-namespace \
 		  --namespace $(DRACON_NS) \
-		  --set "deduplication-db-migrations.image.tag=$(DRACON_VERSION)"
+		  --wait
+	@helm upgrade deduplication-db-migrations ./deploy/deduplication-db-migrations/chart \
```
flowirtz, line 223: this now depends on the draconctl image, which doesn't get automatically published locally, and as such the quickstart guide on a fresh install doesn't work anymore

i.e. running
`make dev-deploy DRACON_VERSION=v0.19.0` on a fresh install crashes with the `draconctl` image not being found

Makefile
```diff
@@ -247,11 +243,12 @@ dev-update-oss-components: cmd/draconctl/bin
...
 		--name $(DRACON_OSS_COMPONENTS_NAME) \
 		./components
-	@helm upgrade $(DRACON_OSS_COMPONENTS_NAME) \
+	helm upgrade $(DRACON_OSS_COMPONENTS_NAME) \
 		./$(DRACON_OSS_COMPONENTS_NAME)-$(DRACON_VERSION).tgz \
 		--install \
 		--namespace dracon \
-		--set container_registry=kind-registry:5000/ocurity/dracon
+		--set container_registry=kind-registry:5000/ocurity/dracon \
+		--values ./deploy/deduplication-db-migrations/values/dev.yaml
```
flowirtz, line 251: why are we passing the `deduplication-db-migrations` to the OSS components? The migrations container isn't even part of this chart?

flowirtz (changes requested):
```
make[1]: Leaving directory '/home/flo/Desktop/Git/dracon'
helm upgrade dracon-oss-components \
	./dracon-oss-components-v0.27.0.tgz \
	--install \
	--namespace dracon \
	--set container_registry=kind-registry:5000/ocurity/dracon \
	--values ./deploy/deduplication-db-migrations/values/dev.yaml
Release "dracon-oss-components" does not exist. Installing it now.
Error: create: failed to create: namespaces "dracon" not found
make: *** [Makefile:242: dev-update-oss-components] Error 1
```

deploy/deduplication-db-migrations/chart/values.yaml
northdpole: these are the default values, this means that its enabled by default, i highly doubt this ever needs to be disabled but good to provide the option

northdpole:
the commits on this PR all modify the Makefile in different ways.
I know this is a hard PR to review so I've added comments to make it easier, I can merge the makefile mods to one commit but it's a lot of small independend changes

Makefile
northdpole: the oss-components have a strong dependency on the migrations since the deduplication db needs the same db values, instead of modifying the draconctl package command to include all of this and thus trying to work around Helm, we share values between the two

Makefile
```diff
@@ -1,16 +1,23 @@
...
 component_containers=$(shell find ./components -name main.go | xargs -I'{}' sh -c 'echo $$(dirname {})/docker')
 component_containers_publish=$(component_containers:docker=publish)
 protos=$(shell find . -not -path './vendor/*' -name '*.proto')
 go_protos=$(protos:.proto=.pb.go)
 latest_tag=$(shell git tag --list --sort="-version:refname" | head -n 1)
 commits_since_latest_tag=$(shell git log --oneline $(latest_tag)..HEAD | wc -l)
-
 GO_TEST_PACKAGES=$(shell go list ./... | grep -v /vendor/)
+
+# Deployment Vars vars
```
flowirtz, line 11: ? proper naming pls

Makefile
```diff
@@ -1,16 +1,23 @@
+# Developper vars
```
flowirtz, line 1: typo but also what does it mean

Makefile
```diff
@@ -210,43 +218,70 @@ deploy-tektoncd-dashboard: tektoncd-dashboard-helm
...
+	@echo "deploying dracon"
 	@helm upgrade dracon ./deploy/dracon/chart \
+	 	  --install \
+		  --values ./deploy/dracon/values/dev.yaml \
+		  --create-namespace \
+		  --set "image.registry=$(CONTAINER_REPO)" \
+		  --namespace $(DRACON_NS) \
+		  --version $(DRACON_VERSION) \
+		  --wait
+	@helm upgrade deduplication-db-migrations ./deploy/deduplication-db-migrations/chart \
```
flowirtz, line 236: would add an echo before this to say "running migrations", just so if the script gets stuck there it's easier to debug

Makefile
```diff
@@ -210,43 +218,70 @@ deploy-tektoncd-dashboard: tektoncd-dashboard-helm
...
+		  --install \
+		  --values ./deploy/deduplication-db-migrations/values/dev.yaml \
+		  --create-namespace \
```
flowirtz, line 239:
```suggestion
```

NS should already exist, from previous command

Makefile
```diff
@@ -210,43 +218,70 @@ deploy-tektoncd-dashboard: tektoncd-dashboard-helm
...
+		  --set "image.registry=$(CONTAINER_REPO)" \
+		  --namespace $(DRACON_NS) \
+		  --set "image.tag=$(DRACON_VERSION)" \
+		  --wait
+
+	# we are seeting the container repo on itself is so we can override the container repo and package url from other make targets
```
flowirtz, line 245: typo and also I don't understand the comment

Makefile
```diff
@@ -210,43 +218,70 @@ deploy-tektoncd-dashboard: tektoncd-dashboard-helm
...
-	@make publish-component-containers CONTAINER_REPO=localhost:5000/ocurity/dracon
+	$(eval CONTAINER_REPO:=localhost:5000/ocurity/dracon)
+
+	$(MAKE) -j 16 publish-component-containers CONTAINER_REPO=$(CONTAINER_REPO)
 	@./bin/cmd/draconctl components package \
 		--version $(DRACON_VERSION) \
 		--chart-version $(DRACON_VERSION) \
 		--name $(DRACON_OSS_COMPONENTS_NAME) \
 		./components
-	@helm upgrade $(DRACON_OSS_COMPONENTS_NAME) \
```
flowirtz, line 244: why are we removing this? Previously I could just run `make dev-update-oss-components` to push the latest dev version into the cluster

flowirtz (changes requested):

Makefile
northdpole: adding a new shortcut for `dev-deploy-oss-components``

Makefile
northdpole: expanding the comment with example

Makefile
northdpole: matter of principle, all helm commands should create the namespace if it doesn't exist

Makefile
flowirtz: ah, I see what you're doing. Thanks for explaining!

I think in that case I'd slightly tweak the target names, to make it more clear what they do:
- `dev-update-oss-components` -> `dev-build-oss-components` maybe?

Also, can we add a convenience method that does the same as before, i.e.
```Makefile
dev-update-oss-components: build-oss-components install-oss-components
```
Otherwise we're making our life harder, from before running one command to now two.

northdpole: convenience method is `dev-deploy-oss-components`` , i'll change the naming on the other one as well

flowirtz (approved):

PR #305, opened 2024-08-12
docs: add CloudPI RFC

albrodfer1:


northdpole:
thanks for the contribution @albrodfer1 ! let me review and will get back to you

PR #306, opened 2024-08-14
Allow overriding image names and make migrations role creation optional

northdpole:


deploy/deduplication-db-migrations/chart/templates/_helpers.tpl
```diff
@@ -60,3 +60,17 @@ Create the name of the service account to use
...
 {{- end }}
 {{- end }}
+
+
+{{/*
+Create the name of the service account to use
+*/}}
+
+{{- define "deduplication_db_migrations.imageName" -}}
+{{- if .Values.deduplication_db_migrations.image.repository }}
```
ptzianos, line 70: this looks more complicated than needed. why not just allow a user to define their repository and make the image be `$registry/$repository:$tag` with defaults for each part?

northdpole: because: ancient ci/cd systems like to set the full image name and path based on their own arbitrary values, so the whole thing needs overriding

newer systems like the one we use allow for more fine grained control so we need to support both at least for now

ptzianos (approved):

PR #308, opened 2024-08-19
close #307 by removing logic which adds duplicate anchors to local tasks

### LINKED ISSUE #307: bug: draconctl adds results anchors twice if the component is referenced locally
```
deployPipeline -> FromReference -> addAnchorResult
deployPipeline->deploymentOrchestrator.Prepare -> components.ProcessTasks -> addAnchorResult
```

example kustomize:

```
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
nameSuffix: -cyberdyne-card-processing
components:
  - pkg:helm/dracon-oss-components/base
  - pkg:helm/dracon-oss-components/git-clone
  - pkg:helm/dracon-oss-components/producer-golang-gosec
  - pkg:helm/dracon-oss-components/producer-golang-nancy
  - pkg:helm/dracon-oss-components/producer-aggregator
  - components/enrichers/knowledgebase
  # - components/enrichers/training
  - pkg:helm/dracon-oss-components/enricher-aggregator
  - pkg:helm/dracon-oss-components/consumer-elasticsearch
  - pkg:helm/dracon-oss-components/consumer-jira
  - pkg:helm/dracon-oss-components/consumer-slack
  # - pkg:helm/dracon-oss-components/consumers/defectdojo

 
```

when running `draconctl pipelines deploy` with the above kustomize only the knowledgebase enricher gets results anchors twice this is because of the duplicate call tree above

northdpole:


ptzianos (changes requested):
ProcessTasks does more than just add the anchor step.
- Why are the other operations done ProcessTassk ok to skip?
- Why wasn't this bug caught by tests?
- And why is this change not tested to prevent future issues like this?
- Why isn't the solution to make the addAnchor* helpers idempotent?

northdpole:
> ProcessTasks does more than just add the anchor step.
> 
> * Why are the other operations done ProcessTassk ok to skip?
> * Why wasn't this bug caught by tests?
> * And why is this change not tested to prevent future issues like this?
> * Why isn't the solution to make the addAnchor* helpers idempotent?

fixed all of these, also made the add env vars idempotent

ptzianos (approved):

PR #319, opened 2024-08-25
Upgrade dt consumer add logging

### LINKED ISSUE #320: Dependency track consumer fails because it can't validate well known certificates

northdpole:
Upgrade the DT consumer library version and adds optional debugging and logging

Closes #320 , #321


components/consumers/dependency-track/main.go
```diff
@@ -47,11 +53,42 @@ func main() {
...
+	}
+	if projectVersion == "" {
+		log.Fatal("project version is mandatory for dependency track")
+	}
+
 	c, err := dtrack.NewClient(authURL, dtrack.WithAPIKey(apiKey))
 	if err != nil {
 		log.Panicf("could not instantiate client err: %#v\n", err)
 	}
+
```
ptzianos, line 73: I'm not a huge fan of this addition, I think it can simplified like this:
```golang
		dtClient, err := dtrack.NewClient(
			authURL,
			dtrack.WithHttpClient(
				 &http.Client{Transport: &http.Transport{
					TLSClientConfig: &tls.Config{
						InsecureSkipVerify: insecure,
					},
				},
			}),
			dtrack.WithDebug(true),
			dtrack.WithAPIKey(apiKey),
		)
		if err != nil {
			log.Panicf("could not instantiate client err: %#v\n", err)
		}
```

components/consumers/dependency-track/main.go
```diff
@@ -164,6 +201,7 @@ func addOwnersTags(owners []string) error {
 }
 
 func uploadBOM(bom string, projectVersion string) (string, error) {
+	slog.Info(fmt.Sprintf("Uploading BOM to Dependency Track for project %s version %s", projectName, projectVersion))
```
ptzianos, line 204: This is not a good use of slog. The logger is expected to receive contextual values as parameters, so that they can be used as filters in a json filtering. The correct use would be like this:
```golang
slog.Info("uploading BOM to Dependency Track, "projectName", projectName, "projectVersion", projectVersion)
```

components/consumers/dependency-track/task.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
 spec:
   description: Pushes findings to a Dependency-Track instance.
   params:
+    # Warning: at the time of writing this api-url is for the port 8081
     - name: consumer-dependency-track-api-url
       type: string
```
dlicheva, line 13: Does it make sense to add the warning to the description? e.g. "Default port is 8081"

components/consumers/dependency-track/task.yaml
```diff
@@ -21,13 +22,18 @@ spec:
...
+      default: "false"
   workspaces:
     - name: output
       description: The workspace containing the source-code to scan.
   steps:
   - name: run-consumer
-    imagePullPolicy: IfNotPresent
-    image: '{{ default "ghcr.io/ocurity/dracon" .Values.image.registry }}/components/consumers/dependency-track:{{ .Chart.AppVersion }}'
+    imagePullPolicy: Always
+    # image: '{{ default "ghcr.io/ocurity/dracon" .Values.image.registry }}/components/consumers/dependency-track:{{ .Chart.AppVersion }}'
```
dlicheva, line 34: Maybe add a ToDO here and a ticket, to have a reminder of where the image is hard-coded.

dlicheva (approved):

components/consumers/dependency-track/task.yaml
northdpole: not sure, this is only valid for local executions of DT, normally you'd not do that

ptzianos: what does this mean? that the API URL is always concatenated with port 8081 if it doesn't have a port?

components/consumers/dependency-track/task.yaml
```diff
@@ -21,12 +22,15 @@ spec:
...
+    - name: consumer-dependency-track-debug
+      type: string
+      default: "false"
   workspaces:
     - name: output
       description: The workspace containing the source-code to scan.
   steps:
   - name: run-consumer
-    imagePullPolicy: IfNotPresent
+    imagePullPolicy: Always
```
ptzianos, line 33: the imagePullPolicy set to Always is only for debugging locally, it should not be set by default

components/consumers/dependency-track/main.go
```diff
@@ -47,11 +53,40 @@ func main() {
...
 	if err != nil {
 		log.Panicf("could not instantiate client err: %#v\n", err)
 	}
-	client = c
+
+	abt, err := client.Metrics.LatestPortfolioMetrics(context.Background())
+	if err != nil {
+		log.Fatalf("cannot connect to Dependency Track at %s, err:'%v'", authURL, err)
+	}
+	slog.Info(fmt.Sprintf("Connection to DT successful, projects in instance: %d", abt.Projects))
```
ptzianos, line 89: you shouldn't use `Sprintf` with slog, there should be a separate field called `projects`

ptzianos:
@northdpole the gosec change has nothing do with this PR and neither the dependency track nor the gosec one have any tickets related to them
also the commit fixing the image should be fixup'd to the commit that introduced the issue

dlicheva (approved):

PR #333, opened 2024-09-02
#332 fix pdf consumer

### LINKED ISSUE #332: pdf consumer does needs to be updated to the new build system

northdpole:
closes #332  by adding a dockerfile and a makefile to the pdf consumer


components/consumers/pdf/main.go
```diff
@@ -99,51 +109,49 @@ func buildPdf(data any) (string, error) {
...
 	}
 
 	reportPage := fmt.Sprintf("file:///%s", reportHTMLPath)
 	if _, err = page.Goto(reportPage); err != nil {
-		return "", fmt.Errorf("could not goto page %s in the browser: %w", reportPage, err)
+		return "", nil, fmt.Errorf("could not goto page %s in the browser: %w", reportPage, err)
 	}
 
-	_, err = page.PDF(playwright.PagePdfOptions{
+	pdfBytes, err := page.PDF(playwright.PagePdfOptions{
```
andream16, line 152: Error check missing here 👀

components/consumers/pdf/Dockerfile
```diff
@@ -1,11 +1,19 @@
-FROM golang:latest
 
-WORKDIR /playwright
-RUN go mod init github.com/ocurity/pdf-consumer &&\
-    go get -u github.com/playwright-community/playwright-go &&\ 
-    go run github.com/playwright-community/playwright-go/cmd/playwright@latest install --with-deps
+# Stage 2: Build
+FROM golang:latest as builder
```
andream16, line 3: Nit: any reason why the images are not pinned?

In my experience it's nicer to pin them so we can gate potential upgrades :)

components/consumers/pdf/Dockerfile
```diff
@@ -1,11 +1,19 @@
...
-ENV PATH="${PATH}:/go/pkg/mod/github.com/playwright-community"
-COPY components/consumers/pdf/pdf /playwright/pdf
-COPY components/consumers/pdf/default.html /playwright/default.html
-ENTRYPOINT ["/playwright/pdf"]
\ No newline at end of file
+# Install playwright cli with right version for later use
+RUN go install github.com/playwright-community/playwright-go/cmd/playwright@latest
+
+# Stage 3: Final
+FROM ubuntu:latest
```
andream16, line 9: Same comment as ⏫  here

components/consumers/pdf/Makefile
```diff
@@ -0,0 +1,18 @@
...
+
+container:
+	cp default.html ../../../bin/components/consumers/pdf/default.html
+	$(DOCKER) build --tag $(CONTAINER_REPO)/components/consumers/pdf:$(DRACON_VERSION) \
+					--file Dockerfile \
+					$$([ "${SOURCE_CODE_REPO}" != "" ] && echo "--label=org.opencontainers.image.source=${SOURCE_CODE_REPO}" ) \
+					 ../../../bin 1>&2
+
+publish:
+	$(DOCKER) push $(CONTAINER_REPO)/components/consumers/pdf:$(DRACON_VERSION) 1>&2
```
ptzianos, line 18: this publish doesn't look like it has a different tag than the one generated automatically

components/consumers/pdf/Dockerfile
```diff
@@ -1,11 +1,19 @@
...
+COPY --from=builder /go/bin/playwright /
+RUN apt-get update && apt-get install -y ca-certificates tzdata \
+    # Install dependencies and all browsers (or specify one)
+    && /playwright install --with-deps \
```
ptzianos, line 13: as far as I understand the code, we are only invoking Chromium in headless mode but this will install other browsers too. why don't we specify just chromium?

components/consumers/pdf/main.go
```diff
@@ -99,51 +109,49 @@ func buildPdf(data any) (string, error) {
...
-		return "", fmt.Errorf("could not apply data to template: %w", err)
+		return "", nil, fmt.Errorf("could not apply data to template: %w", err)
 	}
 
 	pw, err := playwright.Run()
 	if err != nil {
-		return "", fmt.Errorf("could not launch playwright: %w", err)
+		return "", nil, fmt.Errorf("could not launch playwright: %w", err)
 	}
 	defer cleanupRun("could not stop Playwrigh: %w", pw.Stop)
```
ptzianos, line 131: s/Playwrigh/Playwright/

components/consumers/pdf/main.go
```diff
@@ -85,11 +86,20 @@ func sendToS3(filename, bucket, region string) error {
 		return fmt.Errorf("unable to upload %q to %q: %w", filename, bucket, err)
 	}
 
+	_, err = uploader.Upload(&s3manager.UploadInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(strings.Replace(filename, ".html", "", -1) + ".pdf"),
+		Body:   bytes.NewReader(pdfBytes),
+	})
+	if err != nil {
+		return fmt.Errorf("unable to upload %q to %q: %w", filename, bucket, err)
```
ptzianos, line 95: according to Go docs for fmt directives `%q	a single-quoted character literal safely escaped with Go syntax.` . filename and bucket are simple strings though

ptzianos (approved):

components/consumers/pdf/Dockerfile
```diff
@@ -1,11 +1,19 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+COPY components/consumers/pdf/pdf /application/pdf
+COPY components/consumers/pdf/default.html /application/default.html
+WORKDIR /application
+ENTRYPOINT ["/application/pdf"]
```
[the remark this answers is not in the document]
northdpole: it helps make this more understandable, `/playwright` is misleading

components/consumers/pdf/Makefile
northdpole: no it doesn't, i just copied the Makefile from other consumers

components/consumers/pdf/main.go
```diff
@@ -99,51 +109,52 @@ func buildPdf(data any) (string, error) {
...
-		return "", fmt.Errorf("could not goto page %s in the browser: %w", reportPage, err)
+		return "", nil, fmt.Errorf("could not goto page %s in the browser: %w", reportPage, err)
 	}
 
-	_, err = page.PDF(playwright.PagePdfOptions{
+	pdfBytes, err := page.PDF(playwright.PagePdfOptions{
 		Path: playwright.String(reportHTMLPath),
 	})
+	if err != nil {
+		return "", nil, fmt.Errorf("could not generate pdf from page %s, err: %w", reportPage, err)
```
andream16, line 156: ⛏️  I think `err` before wrapping the error is superflous

andream16 (approved):

andream16 (approved):

components/consumers/pdf/main.go
```diff
@@ -82,68 +85,79 @@ func sendToS3(filename, bucket, region string) error {
...
 	return nil
 }
 
-func buildPdf(data any) (string, error) {
+func buildPdf(data any) (string, []byte, error) {
 	tmpl := template.Must(template.ParseFiles("default.html"))
 	cleanupRun := func(msg string, cleanup func() error) {
 		if err := cleanup(); err != nil {
-			log.Printf(msg, err)
+			slog.Error(msg, "error", err)
```
ptzianos, line 109: since you are fixing this, it would be a good idea to join the error with whatever other errors have been generated by the function and return it. you would need to convert the error return value into a named value for this

ptzianos:
@northdpole can't we add some tests in this consumer since we know that it had bugs?

components/consumers/pdf/main.go
```diff
@@ -82,68 +85,79 @@ func sendToS3(filename, bucket, region string) error {
...
 	}
-	defer cleanupRun("could not stop Playwrigh: %w", pw.Stop)
+	defer cleanupRun("could not stop Playwright: %w", pw.Stop)
 
 	browser, err := pw.Chromium.Launch()
 	if err != nil {
-		return "", fmt.Errorf("could not launch Chromium: %w", err)
+		return "", nil, errors.Errorf("could not launch Chromium: %w", err)
 	}
-	defer cleanupRun("could not close browser: %w", browser.Close)
```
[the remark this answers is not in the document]
northdpole: browser.close is needed when you plan on opening multiple browsers multiple times. we are closing playwright and immediately after that stopping the cotnainer, there is no reason to do anything else

northdpole:
> @northdpole can't we add some tests in this consumer since we know that it had bugs?

@ptzianos , i'd need to mock S3 and orchestrate playwright which is a bit of a pain, I'm not saying it'd be a bad idea, but maybe not urgent as part of this pr?

this pr is not fixing bugs, it's making the component work with the make build system.

components/consumers/pdf/Makefile
ptzianos: so it should be removed if its not required?

components/consumers/pdf/main.go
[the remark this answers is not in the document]
ptzianos: under this logic, we shouldn't close file handles, byte streams, connections etc, since everything is running inside a container. but we do clean up things all the time, because you never know how leaving behind all sorts of running resources will interact with other things that you are doing even in short lived workloads. the components are not supposed to have an infinite amount of resources, they should use the minimum things possible and they should keep their environment tidy

ptzianos:
> > @northdpole can't we add some tests in this consumer since we know that it had bugs?
> 
> @ptzianos , i'd need to mock S3 and orchestrate playwright which is a bit of a pain, I'm not saying it'd be a bad idea, but maybe not urgent as part of this pr? also this component has almost no logic, it orchestrates external apis only (load an html document in a browser, print pdf, upload to s3)
> 
> this pr is not fixing bugs, it's making the component work with the make build system.

@northdpole yes you would need to mock the S3 API and then probably build the container in the setup method of the tests and use that to orchestrate playwright and yes, it's going to be a bit of a pain, but this PR is not fixing any bugs because this component couldn't be executed until now, so it's not because it's been running consistently all this time. Even simple looking pieces of code could have bugs in them

components/consumers/pdf/Dockerfile
```diff
@@ -1,11 +1,11 @@
...
+# Install playwright cli with right version for later use
+RUN go install github.com/playwright-community/playwright-go/cmd/playwright@v0.4702.0 && \
+    apt-get update && apt-get install -y ca-certificates tzdata &&\
+    playwright install chromium --with-deps
 
-ENV PATH="${PATH}:/go/pkg/mod/github.com/playwright-community"
-COPY components/consumers/pdf/pdf /playwright/pdf
-COPY components/consumers/pdf/default.html /playwright/default.html
-ENTRYPOINT ["/playwright/pdf"]
+COPY components/consumers/pdf/linux/amd64/pdf /application/pdf
```
andream16, line 8: Can we customise this path based on `GOOS` and `GOARCH` set by the user?

There are a few examples on the repo

components/consumers/pdf/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+CONTAINER_REPO=
+SMITHY_VERSION=
+SOURCE_CODE_REPO=
+PRODUCER_AGGREGATOR_BASE_IMAGE=$(shell test -e .custom_image && cat .custom_image || echo "scratch")
+
+DOCKER=docker
+
+container:
+	$(eval workdir:=$(shell mktemp -d /tmp/pdf.XXXXXX))
```
andream16, line 11: Can we make sure that we're using the correct `--platform` to build the container?

There are a few examples in the repo between Makefile and scripts :)

It would help me running this locally

components/consumers/pdf/main.go
```diff
@@ -8,20 +8,19 @@
...
 	"log"
+	"log/slog"
 	"os"
 	"path/filepath"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/s3/s3manager"
-	"github.com/playwright-community/playwright-go"
+	"github.com/go-errors/errors"
```
andream16, line 19: Can we use `errors` from the standard library?

components/consumers/pdf/main.go
```diff
@@ -52,98 +61,64 @@ func main() {
...
+	slog.Info("reading pdf")
+	resultFilename, pdfBytes, err := buildPdf(responses, pw)
 	if err != nil {
-		return fmt.Errorf("unable to upload %q to %q: %w", filename, bucket, err)
+		return err
 	}
 
-	fmt.Printf("Successfully uploaded %q to %q\n", filename, bucket)
-	return nil
+	slog.Info("uploading pdf to s3", slog.String("filename", resultFilename))
```
andream16, line 95: ⛏️ we can add bucket and region here for visibility

components/consumers/pdf/main.go
```diff
@@ -52,98 +61,64 @@ func main() {
...
+	return s3Wrapper.UpsertFile(resultFilename, bucket, pdfBytes)
 }
 
-func buildPdf(data any) (string, error) {
+func buildPdf(data any, pw playwright.Wrapper) (string, []byte, error) {
 	tmpl := template.Must(template.ParseFiles("default.html"))
```
andream16, line 100: ⛏️ we could use a `template` function that doesn't panic on error so we can handle the error scenario better

pkg/playwright/wrapper.go
```diff
@@ -0,0 +1,58 @@
+package wrapper
+
+import (
+	"github.com/go-errors/errors"
```
andream16, line 4: Can we use `errors` from the standard library?

pkg/s3/wrapper.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"bytes"
+	"log/slog"
+	"os"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/go-errors/errors"
```
andream16, line 12: Same comment here for errors

pkg/s3/wrapper.go
```diff
@@ -0,0 +1,68 @@
...
+	//#nosec:G304
+	data, err := os.ReadFile(htmlFilename) //#nosec:G304
+	if err != nil {
+		return errors.Errorf("could not open file: %w", err)
+	}
+
+	uploader := s3manager.NewUploader(c.session)
+	_, err = uploader.Upload(&s3manager.UploadInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(htmlFilename),
```
andream16, line 48: Should we amend this name by adding a postfix to make it unique? For example a timestamp - just it case it could get overridden

pkg/s3/wrapper.go
```diff
@@ -0,0 +1,68 @@
...
+	slog.Info("uploaded", "filename", htmlFilename, "to", "bucket", bucket, "successfully")
+
+	pdfFilename := strings.Replace(htmlFilename, ".html", "", -1) + ".pdf"
+	_, err = uploader.Upload(&s3manager.UploadInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(pdfFilename),
+		Body:   bytes.NewReader(pdfBytes),
+	})
+	if err != nil {
+		return errors.Errorf("unable to upload %s to %s: %w", pdfBytes, bucket, err)
```
andream16, line 63: We should use `string(pdfBytes)` to print these bytes correctly here :)

pkg/s3/wrapper.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	slog.Info("uploaded", "filename", pdfFilename, "to", "bucket", bucket, "successfully")
```
andream16, line 66: ⛏️ we can use 

```
slog.Info("uploaded successfully", slog.String("filename", pdfFilename), slog.String("bucket", bucket))
```

to leverage log fields and improve searchability

components/consumers/pdf/Makefile
```diff
@@ -0,0 +1,24 @@
...
+BUILD_ARCHITECTURE=
+
+DOCKER=docker
+
+container:
+	$(eval workdir:=$(shell mktemp -d /tmp/pdf.XXXXXX))
+	mkdir -p ${workdir}/components/consumers && \
+	cp -r ../../../bin/components/consumers/pdf ${workdir}/components/consumers && \
+	cp default.html  ${workdir}/components/consumers/pdf && \
+	$(DOCKER) build --tag $(CONTAINER_REPO)/components/consumers/pdf:$(SMITHY_VERSION) \
```
andream16, line 15: Missing `--platform` arg

andream16 (approved):

PR #339, opened 2024-09-05
Feature/ocu 184 allow binaries to be built for architectures different than linux/amd64

andream16:
As per title, allow to customise the built images for 🍎 people and documenting the utilisation.

---

Tested locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,10 @@ BASE_IMAGE=scratch
 DOCKER=docker
 PROTOC=protoc
 
+# https://docs.docker.com/build/building/multi-platform/
+# Make sure to always build containers using AMD64 but allow to be overridden by users if need for cross-os compatibility.
+CONTAINER_ARCH=linux/amd64
+
 export
 
 ########################################
@@ -52,7 +56,7 @@ $(component_binaries):
 component-binaries: $(component_binaries)
 
 $(component_containers): %/docker: %/bin
-	./scripts/build_component_container.sh $@
+	./scripts/build_component_container.sh $@ $(CONTAINER_ARCH)
 
 components: $(component_containers)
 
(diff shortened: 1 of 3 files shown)


ptzianos:
@andream16 yeah I think it would be important to also build the binaries with the correct platform otherwise the containers will fail if we try to run them in a foreign architecture

docs/getting-started.md
```diff
@@ -191,6 +192,41 @@ these components have their own Makefiles. In those cases you can place a
...
+
+###### Containers
+
+If you need your images to be built for non linux/amd64 architecture,
+you can supply the flag `CONTAINER_OS_ARCH` for customisation of containers.
+
+This can be passed to the make commands used to build images, for example:
+
+```bash
+make CONTAINER_OS_ARCH=linux/arm64 components
```
dlicheva, line 207: Optional suggestion: maybe it will be clearer for users if we give an example of exactly how to set it up for your mac. Currently the example just shows the default value, but it might be clearer to say that it might work for mac this way.

dlicheva (approved):

Makefile
```diff
@@ -41,7 +41,11 @@ PROTOC=protoc
 
 # https://docs.docker.com/build/building/multi-platform/
 # Make sure to always build containers using AMD64 but allow to be overridden by users if need for cross-os compatibility.
-CONTAINER_ARCH=linux/amd64
+CONTAINER_OS_ARCH=linux/amd64
```
ptzianos, line 44: this looks like it could just be replaced by `$(GOOS)/$(GOARCH)`

Makefile
```diff
@@ -41,7 +41,11 @@ PROTOC=protoc
...
+# Allow to independently customise go OS and ARCH flags independently and following the standard practices.
+# Defaulting to linux/amd64 as per CONTAINER_OS_ARCH.
+GOOS=linux
```
ptzianos, line 47: this looks like it could cause silent issues for ppl building the binaries to just run them locally. I think it would be better to just set them to `$(shell go env GOOS)` and if someone wants to override them via a environment variable, they can

ptzianos:
@andream16 just a couple of nits as comments, would you also mind cleaning up the history of the PR a little bit. the last commit contains both README fixes and package.json updates and a couple of the initial commits touch the same things again and again because we had a bunch of discussions about this as it was in flight and they can be squashed

andream16:
Declined in favour of https://github.com/ocurity/dracon/pull/343 with better git commits

PR #340, opened 2024-09-06
Adding business logic for reachability enricher - leveraging atom.

andream16:
Code changes to create a new enricher which enriches existing vulnerability reports with [atom](https://github.com/AppThreat/atom) reachability reports.

This is a refactor from https://github.com/ocurity/dracon/pull/271.

I tried to decompose the code by responsibility and added godoc. Let me know if some bits are not clear :).

I opted for passing around concrete types instead of abstractions as we are not leveraging mocks and we're not expecting to swap implementations. This can be done later, if needed. Right now this choice helps with making things clearer and easier to find.

---

Still working on the Task. Will add it later!


andream16:
Need to add task.

andream16:
Need to add README.md

components/enrichers/reachability/cmd/main.go
```diff
@@ -0,0 +1,92 @@
+package main
```
andream16, line 1: If you are not familiar with the concept, code in `internal` [cannot be imported by other go modules](https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface). The build tool would report an error if this happens - leading to a failing build.

This makes sure to leak to potential user only the API that we actually want to expose. This is a nice Go built in feature

components/enrichers/reachability/internal/atom/purl/purl.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Parser allows to extract information from purls - https://github.com/package-url/purl-spec.
+type Parser struct {
+	matcherPurlPkg             *regexp.Regexp
+	matcherPurlTrailingVersion *regexp.Regexp
+	matcherPurlVersion         *regexp.Regexp
+}
+
+func NewParser() (*Parser, error) {
+	purlPkg, err := regexp.Compile(`(?P<p1>[^/:]+/(?P<p2>[^/]+))(?:(?:.|/)v\d+)?@`)
```
northdpole, line 17: why are we using regexp to parse this and not https://github.com/package-url/packageurl-go?

components/enrichers/reachability/internal/atom/reader.go
```diff
@@ -0,0 +1,148 @@
...
+	logger := logging.FromContext(ctx)
+
+	rawPurls, err := jmespath.Search("reachables[].purls[]", reachables)
+	if err != nil {
+		return nil, fmt.Errorf("failed to search reachable purls: %w", err)
+	}
+
+	purls, ok := rawPurls.([]any)
+	if !ok {
+		logger.Error(
```
northdpole, line 104: can we please replace with slog? we are trying to switch components to log in a structured way

components/enrichers/reachability/internal/conf/conf.go
```diff
@@ -0,0 +1,59 @@
+package conf
```
northdpole, line 1: perhaps this should be part of the base enricher?

components/enrichers/reachability/internal/atom/purl/purl.go
andream16: Will look into it!

components/enrichers/reachability/internal/fs/fs.go
```diff
@@ -0,0 +1,97 @@
+package fs
```
northdpole, line 1: perhaps this should also be part of the base enricher. not for this PR but maybe good to be in a ticket for the OSS refactoring project

components/enrichers/reachability/internal/search/search.go
```diff
@@ -0,0 +1,108 @@
...
+// NewSearcher returns a new searcher.
+func NewSearcher(reachables atom.Reachables, reachablePurls atom.ReachablePurls) (*Searcher, error) {
+	switch {
+	case len(reachables) == 0:
+		return nil, errors.New("invalid empty reachables")
+	case len(reachablePurls) == 0:
+		return nil, errors.New("invalid empty reachable purls")
+	}
+
+	matcherFileLine, err := regexp.Compile(`(?P<file>[^/]+):(?P<line>[\d-]+)`)
```
northdpole, line 29: does atom only have lines? 
i thought it had ranges, happy to be wrong if your tests show otherwise :)

components/enrichers/reachability/internal/search/search.go
```diff
@@ -0,0 +1,108 @@
...
+	// Otherwise check for a file match
+	if match := s.matcherFileLine.FindStringSubmatch(search); len(match) > 0 {
+		var (
+			file        = match[1]
+			fileContent = match[2]
+		)
+
+		lineStart, lineEnd, err := s.searchLineRange(fileContent)
+		if err != nil {
+			return false, fmt.Errorf("failed to parse line range: %w", err)
```
northdpole, line 61: we use go-errors Errorf by convention

andream16:
Closed in favour of https://github.com/ocurity/dracon/pull/360#pullrequestreview-2310179723

components/enrichers/reachability/internal/search/search.go
andream16: From my tests and my understanding it looks ok but please feel free to test too!

components/enrichers/reachability/internal/conf/conf.go
andream16: What do you mean by that?

PR #342, opened 2024-09-10
fix-bug-341-postgreql-does-not-have-credentials

### LINKED ISSUE #341: Postgresql password is not set during make install anymore
A change in the last two weeks has removed the postgresql password. As a result postgres generates its own and thus no migrations or components can be run

northdpole:
closes #341 
A change in the last two weeks has removed the postgresql password. As a result postgres generates its own and thus no migrations or components can be run

### DIFF
--- a/deploy/dracon/chart/values.yaml
+++ b/deploy/dracon/chart/values.yaml
@@ -38,9 +38,10 @@ arangodb:
     className: ""
     host: ""
 
-postgresql:
-  # if set, a PostgreSQL instance will be deployed
-  enabled: true
+# this section controls the database that components have access to
+# the database should use the Postgres dialect.
+# the database is used by the deduplication enricher
+database: &psqlConfig
   host: ""
   auth:
     username: ""
@@ -50,19 +51,20 @@ postgresql:
     querystringargs: ""
   fullnameOverride: ""
 
+postgresql:
+  # if set, a PostgreSQL instance will be deployed
+  enabled: true
+  <<: *psqlConfig
+
+
 # this section controls aspects of managing a database used to store deduplication enrichments
 # the database should use the Postgres dialect.
 deduplication-db-migrations:
   # if set, a Job will be deployed that applies migrations to the deduplication database
   # the Job will run as part of the post-install/post-upgrade hook
   enabled: true
-  database:
-    host: ""
-    auth:
-      username: ""
-      password: ""
-      database: ""
-      querystringargs: ""
+  database: *psqlConfig
+
 
 image:
   # registry to use for all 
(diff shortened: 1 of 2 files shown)


deploy/dracon/chart/values.yaml
```diff
@@ -56,13 +56,12 @@ deduplication-db-migrations:
...
   # the Job will run as part of the post-install/post-upgrade hook
   enabled: true
-  database:
-    host: ""
-    auth:
-      username: ""
-      password: ""
-      database: ""
-      querystringargs: ""
+  database: *psqlConfig
```
ptzianos, line 59: this looks odd, the postgresql section should accept an auth section, not a database section https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L32

ptzianos: why not use the anchor for the auth dictionary of the database and pass that to the auth section of the postgresql configuration?

northdpole: `auth` does not include the `host` attribute however

andream16 (approved):

deploy/dracon/chart/values.yaml
ptzianos:
```yaml

# this section controls the database that components have access to
# the database should use the Postgres dialect.
# the database is used by the deduplication enricher
database: &psqlConfig
  host: "aaaaaaa"
  auth:
    username: "bla"
    password: ""
    database: ""
    postgresPassword: ""
    querystringargs: ""

postgresql:
  # if set, a PostgreSQL instance will be deployed
  enabled: true
  <<: *psqlConfig

# this section controls aspects of managing a database used to store deduplication enrichments
# the database should use the Postgres dialect.
deduplication-db-migrations:
  # if set, a Job will be deployed that applies migrations to the deduplication database
  # the Job will run as part of the post-install/post-upgrade hook
  enabled: true
  database: *psqlConfig

image:
  # registry to use for all 
  registry: ""
```
after offline discussions

deploy/dracon/values/dev.yaml
```diff
@@ -30,28 +30,22 @@ arangodb:
...
     username: dracon
     password: dracon
     database: dracon
     postgresPassword: dracon
     querystringargs: "sslmode=disable"
-  host: dracon-postgresql:5432
+
+postgresql:
+  enabled: true
+  <<: *&psqlConfig
```
ptzianos, line 44: this doesn't look correct

ptzianos (approved):

ptzianos (approved):

PR #343, opened 2024-09-11
Feature/ocu 184 allow binaries and containers to be built for architectures different than amd

andream16:
As per title, allow to customise the built images and binaries for 🍎 people and documenting the utilisation.

We basically use [--platform](https://docs.docker.com/build/building/multi-platform/) when building to specify for which platform we want to build an image. At the same time we can customise for which architecture the Go binaries should be built for.

---

Based on the supplied ARCH and OS, a binary is created in a self-documented sub-folder, as follows:

Screenshot 2024-09-09 at 15 04 16

---

It would be nice to collapse the different `Dockerfile`s and `Makefile`s in components into a single one, so such customisation can be centralised. I'll start a thread in the chat about it.

---

Tested locally:
- `make components` -> builds with `linux/amd64` by default. Running the containers works fine.
- `make GOOS=linux GOARCH=arm64 CONTAINER_OS_ARCH=linux/arm64 components` -> builds with `linux/arm64`. Running the containers work.
- `make GOOS=darwin GOARCH=arm64 component-binaries` -> build with `darwin/arm64` and I can exec the binaries fine on my machine.

---

I would appreciate if you can try these commands yourself and if you have other pointers to what to test.

--- 

Minor: bumping `remark-cli` to v12.0.1 to suppress formatting errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,14 @@ BASE_IMAGE=scratch
 DOCKER=docker
 PROTOC=protoc
 
+# https://docs.docker.com/build/building/multi-platform/
+# Make sure to always build containers using AMD64 but allow to be overridden by users if need for cross-os compatibility.
+CONTAINER_OS_ARCH=linux/amd64
+# Allow to independently customise go OS and ARCH flags.
+# Defaulting to linux/amd64 as per CONTAINER_OS_ARCH.
+GOOS=linux
+GOARCH=amd64
+
 export
 
 ########################################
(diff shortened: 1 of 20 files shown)


scripts/build_component_container.sh
```diff
@@ -4,35 +4,52 @@ set -e;
...
-echo "${1}" | grep -Eq ^components/enrichers/.*$ && executable="${executable}" || true
-echo "${1}" | grep -Eq ^components/consumers/.*$ && executable="${executable}" || true
+echo ${dir_name} | grep -Eq ^components/producers/.*$ && executable="${executable}-parser" || true
+echo ${dir_name} | grep -Eq ^components/enrichers/.*$ && executable="${executable}" || true
+echo ${dir_name} | grep -Eq ^components/consumers/.*$ && executable="${executable}" || true
 
-executable_src_path=$(dirname "${1}")
-executable_path=$(dirname "${1}")/"${executable}"
+executable_src_path=$(dirname $dir_name)
+executable_path=$(dirname $dir_name)/${build_architecture}/"${executable}"
```
ptzianos, line 36: nit: can you encapsulate please the BASH variables in curly brackets here and here: https://github.com/ocurity/dracon/pull/343/commits/eea40403dc9292ecd571b7b09613b41b1615a2f2#diff-e886250d151c99b2b701ddeab8e0e58bcac8f434e02ead4aafaab12462fa8e86R25 for consistency?

Makefile
```diff
@@ -47,22 +55,25 @@ export
...
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o bin/cmd/$(GOOS)/$(GOARCH)/draconctl cmd/draconctl/main.go
 
 draconctl-image: cmd/draconctl/bin
 	$(DOCKER) build -t "${CONTAINER_REPO}/draconctl:${DRACON_VERSION}" \
+		--build-arg GOOS=$(GOOS) \
+		--build-arg GOARCH=$(GOARCH) \
 		$$([ "${SOURCE_CODE_REPO}" != "" ] && echo "--label=org.opencontainers.image.source=${SOURCE_CODE_REPO}" ) \
-		-f containers/Dockerfile.draconctl .
+		-f containers/Dockerfile.draconctl . \
+		--platform=$(CONTAINER_OS_ARCH)
```
ptzianos, line 76: as far as I can see this is the only point where the CONTAINER_OS_ARCH is used. can we replace it with just `$(GOOS)/$(GOARCH)`

andream16:
Closed in favour of https://github.com/ocurity/dracon/pull/350/files

PR #345, opened 2024-09-12
Bugfix/344 add certificates to consumers

### LINKED ISSUE #344: Base Image does not have certificates
the following consumers are missing certificates in their dockerfiles thus cannot work in a production environment we have many components that need certificates so let's add the certs in the base image

* defectdojo
* arangodb
* aws-s3
* bigquery
* elasticsearch
* mongodb

northdpole:
close #344 by adding certificates to consumers


components/consumers/elasticsearch/Makefile
```diff
@@ -0,0 +1,17 @@
+.PHONY: container publish
+
+CONTAINER_REPO=
+DRACON_VERSION=
+SOURCE_CODE_REPO=
+PRODUCER_AGGREGATOR_BASE_IMAGE=$(shell test -e .custom_image && cat .custom_image || echo "scratch")
```
ptzianos, line 6: this name is wrong

components/consumers/elasticsearch/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+DOCKER=docker
+
+container:
+	$(DOCKER) build --tag $(CONTAINER_REPO)/components/consumers/elasticsearch:$(DRACON_VERSION) \
+					--file Dockerfile \
+					$$([ "${SOURCE_CODE_REPO}" != "" ] && echo "--label=org.opencontainers.image.source=${SOURCE_CODE_REPO}" ) \
+					 ../../../bin 1>&2
+
+publish:
```
ptzianos, line 16: the publish target seems to be doing what the default on is doing

components/consumers/mongodb/Makefile
```diff
@@ -0,0 +1,17 @@
+.PHONY: container publish
+
+CONTAINER_REPO=
+DRACON_VERSION=
+SOURCE_CODE_REPO=
+PRODUCER_AGGREGATOR_BASE_IMAGE=$(shell test -e .custom_image && cat .custom_image || echo "scratch")
```
ptzianos, line 6: this name is wrong

components/consumers/mongodb/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+DOCKER=docker
+
+container:
+	$(DOCKER) build --tag $(CONTAINER_REPO)/components/consumers/mongodb:$(DRACON_VERSION) \
+					--file Dockerfile \
+					$$([ "${SOURCE_CODE_REPO}" != "" ] && echo "--label=org.opencontainers.image.source=${SOURCE_CODE_REPO}" ) \
+					 ../../../bin 1>&2
+
+publish:
```
ptzianos, line 16: the publish target seems to be doing what the default on is doing

containers/Dockerfile.base
```diff
@@ -0,0 +1,16 @@
...
+    apk add --no-cache ca-certificates && \
+    update-ca-certificates
+
+FROM scratch
+
+ARG executable_path
+
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+COPY ${executable_path} /app/${executable_path}
```
andream16, line 15: ❤️

containers/Dockerfile.base
```diff
@@ -0,0 +1,16 @@
+FROM golang:alpine AS builder
+
```
andream16, line 2: ⛏️ there's an extra newline here

andream16 (approved):

scripts/build_component_container.sh
```diff
@@ -23,14 +24,22 @@ if make -C "${executable_src_path}" --no-print-directory --dry-run container >/d
 then
     make -C "${executable_src_path}" --no-print-directory --quiet container CONTAINER_REPO="${CONTAINER_REPO}" DRACON_VERSION="${DRACON_VERSION}"
 else
-    dockerfile_template="
-        FROM ${BASE_IMAGE:-scratch}                     \n
-        COPY ${executable_path} /app/${executable_path} \n
-        ENTRYPOINT ["/app/${executable_path}"]          \n
-    "
+    dockerfile_template=""
```
ptzianos, line 27: I have a lot of reservations about these changes. this looks like it can be simplified to just:
1. I don't see why there should be 2 Dockerfiles, we can just have one based on alpine and if the user wants a different base to wrap all the executables they can supply a different `BASE_IMAGE_PATH`. Which would simplify this to:
```bash
BASE_IMAGE_PATH=${BASE_IMAGE_PATH:-./containers/Dockerfile.base}
docker build \
        --build-arg executable_path=${executable_path}\
        -t "${CONTAINER_REPO}/${executable_src_path}:${DRACON_VERSION}" \
        $([ "${SOURCE_CODE_REPO}" != "" ] && echo "--label=org.opencontainers.image.source=${SOURCE_CODE_REPO}" ) \
        -f "${BASE_IMAGE_PATH}" ./bin
```

Also, it's customary for shell environment variables to be uppercase, so I think the `executable_path` in the Dockerfile should be uppercase

This should simplify the whole PR quite a bit

components/producers/ossf-scorecard/Dockerfile
```diff
@@ -1,8 +1,3 @@
-ARG OSSF_SCORECARD_SAFETY_BASE_IMAGE
 FROM gcr.io/openssf/scorecard:stable
 
-FROM ${OSSF_SCORECARD_SAFETY_BASE_IMAGE}
-
-COPY --from=0 /scorecard /scorecard
-
-ENTRYPOINT ["/scorecard"]
\ No newline at end of file
+ENTRYPOINT ["/scorecard"]
```
ptzianos, line 3: why is this needed any more? isn't the original image just invoking the `scorecard` binary?

scripts/build_component_container.sh
```diff
@@ -23,14 +24,22 @@ if make -C "${executable_src_path}" --no-print-directory --dry-run container >/d
...
+    if [[ -v "${BASE_IMAGE+x}" && -n "${BASE_IMAGE+x}" ]]
```
andream16, line 28: On my machine this fails with error `conditional binary operator expected`

I think this can simplified by just using `if [[ -n "${BASE_IMAGE+x}" ]]` as `-n` checks already for `BASE_IMAGE` to be set.

Works on my machine like this!

components/producers/ossf-scorecard/Makefile
```diff
@@ -2,7 +2,7 @@
 
 CONTAINER_REPO=
 DRACON_VERSION=
-OSSF_SCORECARD_SAFETY_BASE_IMAGE=$(shell test -e .custom_image && cat .custom_image || echo "alpine:latest")
+OSSF_SCORECARD_SAFETY_BASE_IMAGE="alpine:latest"
```
ptzianos, line 5: I don't think we need this at all tbh. We can parameterise the image path in the task and if a user wants a different image with scorecard they can set that in their pipeline runs, same way we do it in the git-clone task

andream16 (dismissed):
Left a comment for the Makefile.

I also tried to build the containers using `make components` but I get the following error when running one of them:

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/app/${executable_path}": stat /app/${executable_path}: no such file or directory: unknown.
```

It seems like `executable_path` is not getting interpolated correctly 🤔

scripts/build_component_container.sh
```diff
@@ -23,14 +24,22 @@ if make -C "${executable_src_path}" --no-print-directory --dry-run container >/d
...
+    else
+        BASE_IMAGE=''
+        dockerfile_template=$(cat "${containers_path}/Dockerfile.base")
+    fi
+
     dockerfile_path=$(mktemp)
     printf "${dockerfile_template}" > "${dockerfile_path}"
-    docker build -t "${CONTAINER_REPO}/${executable_src_path}:${DRACON_VERSION}" \
+    docker build \
+        --build-arg executable_path=${executable_path}\
```
andream16, line 40: ⛏️ can we rename this argument to uppercase to comply with the other argument?

containers/Dockerfile.base
```diff
@@ -0,0 +1,15 @@
+FROM golang:alpine AS builder
```
ptzianos, line 1: we should pin this to the alpine based image with the Go version we are using

components/producers/github-code-scanning/Makefile
```diff
@@ -12,6 +11,3 @@ container:
 					--file Dockerfile \
 					$$([ "${SOURCE_CODE_REPO}" != "" ] && echo "--label=org.opencontainers.image.source=${SOURCE_CODE_REPO}" ) \
 					 ../../../bin 1>&2
-
-publish:
-	$(DOCKER) push $(CONTAINER_REPO)/components/producers/github-code-scanning:$(DRACON_VERSION) 1>&2
```
ptzianos, line 17: The Dockerfile doesn't seem to be doing anything other than wrapping the executable and adding certificates in it

ptzianos (approved):

PR #355, opened 2024-09-16
Making sure that the example make publish-component-containers highli…

andream16:
Tiny little detail that is important to get the demo to work locally and avoid pods going in ImagePullBackoff as the images that they are looking to pull are not found.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -165,7 +165,7 @@ cluster, but make sure that you replace the URL with the registry URL that you
 are using, if you are using something else:
 
 ```bash
-make publish-component-containers CONTAINER_REPO=localhost:5000/ocurity/dracon
+make publish-component-containers CONTAINER_REPO=localhost:5000/ocurity/dracon DRACON_VERSION=0.39.0
 ```
 
 *\*Notice that the repo we are using is slightly different from the


docs/getting-started.md
```diff
@@ -165,7 +165,7 @@ cluster, but make sure that you replace the URL with the registry URL that you
 are using, if you are using something else:
 
 ```bash
-make publish-component-containers CONTAINER_REPO=localhost:5000/ocurity/dracon
+make publish-component-containers CONTAINER_REPO=localhost:5000/ocurity/dracon DRACON_VERSION=0.39.0
```
northdpole, line 168: this means that we need to update the docs with every release, extra manual effort that i'm not super excited about.

doesn't this default to an automatically-calculated version?

dlicheva: @ptzianos added something similar here: https://github.com/ocurity/dracon-saas/pull/616/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52 
The makefile refers to the latest version, so the specific version is not included in the readme

ptzianos: I would say that it would be a good idea to document the semantics of this. if you supply the DRACON_VERSION, then all components will be tagged as that DRACON_VERSION and if a version like that already exists and is published but you have done changes locally, you might have different behaviour than what you expect. If the environment variable is not supplied, one will be automatically calculated based on the latest tag which should make it clear that you are using something that is modified based on your local changes

andream16: Makes sense, I can look into making sure that DRACON_VERSION is used  based on latest and can remove the mentions of a specific version from the getting started docs. Right now we need to use DRACON_VERSION=XXXX in a few spots

PR #360, opened 2024-09-17
Feature/reachability enricher

andream16:
Code changes to create a new enricher which enriches existing vulnerability reports with [atom](https://github.com/AppThreat/atom) reachability reports.

This is a refactor from https://github.com/ocurity/dracon/pull/271.

I tried to decompose the code by responsibility and added godoc. Let me know if some bits are not clear :).

I opted for passing around concrete types instead of abstractions as we are not leveraging mocks and we're not expecting to swap implementations. This can be done later, if needed. Right now this choice helps with making things clearer and easier to find.

---

Tested the task locally with a custom pipeline and works fine!

---

With @northdpole we decided to not re-use the existing `cdxgen` producer as it's more "generic" and our step is more specialised instead - for example, it needs a specific programming language flag to be passed.

We can refactor this Task  in future if needed.


components/enrichers/reachability/internal/atom/purl/purl.go
```diff
@@ -0,0 +1,143 @@
...
+	return &Parser{
+		matcherPurlPkg:             purlPkg,
+		matcherPurlTrailingVersion: purlTrailingVersion,
+		matcherPurlVersion:         purlVersion,
+	}, nil
+}
+
+// ParsePurl extracts pkg:version matches from the supplied purl.
+func (p *Parser) ParsePurl(purl string) ([]string, error) {
+	purl = p.matcherPurlTrailingVersion.ReplaceAllString(purl, "$1@")
```
andream16, line 39: I'm refactoring this to use `packageurl` but need some extra time to make sure that the refactor actually makes sense

From my findings, the actual code spits out the following:

```go
func TestParser_ParsePurl(t *testing.T) {
	pParser, err := purl.NewParser()
	require.NoError(t, err)

	for _, tt := range []struct {
		inputPurl       string
		expectedMatches []string
	}{
		{
			inputPurl: "pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c",
			expectedMatches: []string{
				"birkenfeld/pygments-main:244",
				"birkenfeld/pygments-main:244fd47e07d1014f0aed9c",
				"pygments-main:244",
				"pygments-main:244fd47e07d1014f0aed9c",
			},
		},
	} {
		t.Run(fmt.Sprintf("it should correctly parse with input %s", tt.inputPurl), func(t *testing.T) {
			pp, err := pParser.ParsePurl(tt.inputPurl)
			require.NoError(t, err)
			assert.Equal(t, pp, tt.expectedMatches)
		})
	}
}
```

And I was able to refactor is as:

```go
func (p *Parser) ParsePurl(purl string) ([]string, error) {
	pp, err := packageurl.FromString(purl)
	if err != nil {
		return nil, fmt.Errorf("failed to parse purl: %w", err)
	}

	var (
		namespace    = pp.Namespace
		name         = pp.Name
		version      = pp.Version
		shortVersion string
	)

	if len(version) >= 3 {
		shortVersion = version[:3]
	}

	return []string{
		path.Join(namespace, name) + ":" + shortVersion,
		path.Join(namespace, name),
	}, nil
}
```

---

To me the arbitrary choices of extracting partial versions are bizarre - will try to make more sense of them

Feel free to share your thoughts on it @northdpole

northdpole: they are bizare indeed, i think it's the specific case where the version is a commit sha however

components/enrichers/reachability/main_test.go
```diff
@@ -0,0 +1,102 @@
...
+			expFilePath := getExpectedPath(t, fp)
+			require.FileExistsf(t, resFilePath, "expected file %s doesn't exist in path", expFilePath)
+
+			resFile, err := os.ReadFile(resFilePath)
+			require.NoErrorf(t, err, "could not open results file %s", resFilePath)
+
+			expFile, err := os.ReadFile(expFilePath)
+			require.NoErrorf(t, err, "could not open expectations file %s", expFilePath)
+
+			assert.Equalf(
```
northdpole, line 82: we use `require` instead of assert, just an internal convention

andream16: From my experience, `require` is good to be used when we want to terminate a test if an assertion fails.

It's a good use case for when, for example, we want to make sure that a slice has 3 elements and then after we access element 0, 1 and 2 to make some assertions. Exiting earlier in that case makes sense to avoid panics.

On the other end, `assert` makes sense to still fail a test but don't stop it. In this way we can also perform other assertions later. This is usually better as the feedback loop to fix multiple tests is faster imo. Otherwise, with require, I could only see one failed test at a time!

components/enrichers/reachability/internal/atom/purl/purl.go
andream16: I'll raise a smaller PR to improve this bit, if that's ok!

PR #363, opened 2024-09-18
Feature/362 example snyk docker producer

### LINKED ISSUE #362: create an example snyk producer
Snyk-cli has several containers one for each language they support, until we get feature requests for more let's create a producer for the dogfooding one for Go at least.

### LINKED ISSUE #364: bug: if during parsing of sarif input, there is no endLine for a finding, the target is malformed
the target is e.g. `package.json:1-` instead of the convention `package.json:1-1`

northdpole:
closes #362 and to create has to fix #364 

 can be tested with 

create an account at snyk.io, obtain an API key, paste your api key in ./examples/pipelines/snyk-project/pipelinerun.yaml as the value of the parameter `producer-snyk-docker-api-key`
then run

```
make dev-teardown install # cleanup
bin/cmd/draconctl pipelines deploy ./examples/pipelines/snyk-project # setup the workflow
kubectl -n dracon create -f ./examples/pipelines/snyk-project/pipelinerun.yaml
```

you can monitor the workflow with 
`kubectl -n dracon get po -w`
once the consumer `stdout-json` finishes you can see the findings in the logs


components/producers/producer.go
```diff
@@ -123,6 +123,7 @@ func WriteDraconOut(
 
 	stat, err := os.Stat(OutFile)
 	if Append && err == nil && stat.Size() > 0 {
+		slog.Info("appending", "issues", len(cleanIssues), "tool", toolName, "to", OutFile)
```
andream16, line 126: You can leverage log fields for this and the tuples key:val would be more easy to understand imo

```
slog.Info(
    "appending", 
    slog.String("issues", len(cleanIssues)), 
    slog.String("tool", toolName), 
    slog.String("to", OutFile),
)
```

components/producers/snyk-docker/main.go
```diff
@@ -0,0 +1,55 @@
...
+		log.Fatal(err)
+	}
+	if err := writeOutput(results); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func writeOutput(results map[string][]*v1.Issue) error {
+	for _, issues := range results {
+		slog.Info("writing", "issues", len(issues), "tool", "snyk")
```
andream16, line 34: Same comment here as in `producer.go`

components/producers/snyk-docker/main.go
```diff
@@ -0,0 +1,55 @@
...
+		if err := producers.WriteDraconOut(
+			"snyk",
+			issues,
+		); err != nil {
+			return err
```
andream16, line 39: Should we log and continue here instead of failing the whole task?

components/producers/snyk-docker/task.yaml
```diff
@@ -0,0 +1,58 @@
...
+        echo "Snyk completed successfully! exitcode $exitCode"
+      fi
+    volumeMounts:
+    - mountPath: /scratch
+      name: scratch
+
+  - name: produce-issues
+    imagePullPolicy: Always
+    # image: '{{ default "ghcr.io/ocurity/dracon" .Values.image.registry }}/components/producers/snyk:{{ .Chart.AppVersion }}'
+    image: 'kind-registry:5000/ocurity/dracon/components/producers/snyk:v0.39.0-68faeadd-dirty'
```
andream16, line 51: We probably need to commit the change on line :50 here 🤓

northdpole: good catch!

andream16 (approved):

PR #365, opened 2024-09-19
Simplifying purl parsing and reachability flow for atom reachability …

andream16:
Simplified how we extract sub-parts to be used for reachability analysis in pURLS. Tested that we interpret correctly both SEMVER and SHA commits (with correct short version extraction for SHA).

Also simplified reachability flow.

---

All the existing tests pass showing that this works as expected :)

### DIFF
--- a/components/enrichers/reachability/internal/atom/reader.go
+++ b/components/enrichers/reachability/internal/atom/reader.go
@@ -9,8 +9,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/jmespath/go-jmespath"
-
 	"github.com/ocurity/dracon/components/enrichers/reachability/internal/atom/purl"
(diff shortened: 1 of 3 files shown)


components/enrichers/reachability/internal/atom/reader.go
```diff
@@ -94,49 +92,29 @@ func (r *Reader) Read(ctx context.Context) (*Response, error) {
 
 // ReachablePurls finds all the reachable purls presents in the atom reachability result.
 func (r *Reader) ReachablePurls(ctx context.Context, reachables *Response) (ReachablePurls, error) {
-	logger := logging.FromContext(ctx)
-
-	rawPurls, err := jmespath.Search("reachables[].purls[]", reachables)
```
andream16, line 99: I removed this dependency here as it's not really needed and can be simplified.

northdpole (approved):

PR #369, opened 2024-09-19
Feature/356 checkov producer

### LINKED ISSUE #356: Create a producer for checkov

northdpole:
close bug #367 
close #356 create a producer for checkov

test with

```
make install
make publish-containers CONTAINER_REPO=localhost:5000/ocurity/dracon    
```

in checkov's `task.yaml` replace the image with the published image
 then run

```
bin/cmd/draconctl pipelines deploy ./examples/pipelines/checkov-project         
kubectl -n dracon create -f ./examples/pipelines/checkov-project/pipelinerun.yaml     
```


components/producers/checkov/main.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	if err := run(inFile); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run(inFile []byte) error {
+	sarifResults, sarifErr := handleSarif(inFile)
+	cyclondxResults, cyclonedxErr := handleCycloneDX(inFile)
+	if sarifErr != nil && cyclonedxErr != nil {
```
ptzianos, line 38: shouldn't you be checking if the sarif parsing worked and then skip the cyclonedx parsing if it did?

components/producers/checkov/main.go
```diff
@@ -0,0 +1,72 @@
...
+		return errors.Errorf("Could not parse input file as neither Sarif nor CycloneDX sarif error: %v, cyclonedx error: %v", sarifErr, cyclonedxErr)
+	}
+
+	var issues []*v1.Issue
+	if sarifResults != nil {
```
ptzianos, line 43: this if section could be merged into a single if section in the beginning of the function where we first check if the sarif parsing worked and then skip cycloneDX partsing

components/producers/checkov/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	v1 "github.com/ocurity/dracon/api/proto/v1"
+	"github.com/ocurity/dracon/pkg/cyclonedx"
+	"github.com/ocurity/dracon/pkg/sarif"
+
+	"github.com/ocurity/dracon/components/producers"
+)
+
+// the CycloneDX target override
+var Target string
```
ptzianos, line 17: why does the Target need to be a public variable? also, this pattern is very easy to cause issues and weird interactions between tests, it should be passed as a parameter in a simpler manner not as a global variable

components/producers/checkov/main.go
```diff
@@ -0,0 +1,72 @@
+package main
+
+import (
+	"flag"
+	"log"
+
+	"github.com/go-errors/errors"
```
ptzianos, line 7: please cleanup the imports, the standard library imports should be first, then the third party ones and finally the imports coming from this repository

components/producers/checkov/main_test.go
```diff
@@ -0,0 +1,105 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"google.golang.org/protobuf/proto"
```
ptzianos, line 9: same for these imports

components/producers/checkov/main_test.go
```diff
@@ -0,0 +1,105 @@
...
+			Target:      "code/cfngoat/.github/workflows/main.yaml:1-1",
+			Type:        "CKV2_GHA_1",
+			Title:       "Ensure top-level permissions are not set to write-all",
+			Severity:    v1.Severity_SEVERITY_HIGH,
+			Description: "MatchedRule: {\"id\":\"CKV2_GHA_1\",\"name\":\"Ensure top-level permissions are not set to write-all\",\"shortDescription\":{\"text\":\"Ensure top-level permissions are not set to write-all\"},\"fullDescription\":{\"text\":\"Ensure top-level permissions are not set to write-all\"},\"defaultConfiguration\":{\"level\":\"error\"},\"help\":{\"text\":\"Ensure top-level permissions are not set to write-all\\nResource: on(build)\"}} \n Message: Ensure top-level permissions are not set to write-all",
+		},
+	}
+
+	foundIssues := 0
+	for _, wroteIssue := range wrote.Issues {
```
ptzianos, line 63: this looks like it could be simplified with a sort on the target in both slices and then use the EqualValues function

pkg/cyclonedx/cyclonedx_test.go
```diff
@@ -66,3 +66,31 @@ func TestToDraconSaaSInfra(t *testing.T) {
...
+	expectedIssues := []*v1.Issue{
+		{
+			Target:        "my-awesome-infra",
+			Type:          "SBOM",
+			Title:         "SBOM for my-awesome-infra",
+			Severity:      v1.Severity_SEVERITY_INFO,
+			CycloneDXSBOM: &saasBOM,
+		},
+	}
+	assert.Equal(t, expectedIssues[0].Target, issues[0].Target)
```
ptzianos, line 87: we should use require instead of assert to immediately terminate the tests with a message about what failed

examples/pipelines/checkov-project/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+nameSuffix: -checkov-project
+components:
+  - pkg:helm/dracon-oss-components/base
+  - pkg:helm/dracon-oss-components/git-clone
+  - components/producers/checkov
```
ptzianos, line 8: this is ok when creating the component locally, however it will cause issues when the PR is merged and a release is created. it should be fixed but just before merging

components/producers/checkov/main_test.go
northdpole: protos have a ton of fields that are conflicting, I only want to test for equality on what matters

ptzianos (approved):

dlicheva (approved):

PR #380, opened 2024-09-26
fix #379, add significantly more error logging

### LINKED ISSUE #379: Elasticsearch consumer needs more helpful error logging
e.g. if the version cannot be retrieved due to incorrect permissions on the api key it just crashes instead of showing a nice error message<br><br>same for non-200 response when posting things to the index or if the index doesn't exist

northdpole:
closes #379 
adds logs such as:
```{"time":"2024-09-26T11:57:01.590757103Z","level":"ERROR","msg":"could not contact remote Elasti
csearch: ","scanID":"","error":"could not contact Elasticsearch, attempted to retrieve cluster 
info and got status code: 401 as a result, body: [401 Unauthorized] {\"error\":{\"root_cause\":
[{\"type\":\"security_exception\",\"reason\":\"unable to authenticate with provided credentials
 and anonymous access is not allowed for this request\",\"additional_unsuccessful_credentials\"
:\"API key: unable to find apikey with id kkttHpIBE8LvqBqM9yuN\",\"header\":{\"WWW-Authenticate
\":[\"Basic realm=\\\"security\\\", charset=\\\"UTF-8\\\"\",\"Bearer realm=\\\"security\\\"\",\
"ApiKey\"]}}],\"type\":\"security_exception\",\"reason\":\"unable to authenticate with provided
 credentials and anonymous access is not allowed for this request\",\"additional_unsuccessful_c
redentials\":\"API key: unable to find apikey with id kkttHpIBE8LvqBqM9yuN\",\"header\":{\"WWW-
Authenticate\":[\"Basic realm=\\\"security\\\", charset=\\\"UTF-8\\\"\",\"Bearer realm=\\\"secu
rity\\\"\",\"ApiKey\"]}},\"status\":401}"}```

when the api key cannot contact elasticsearch at all

and similar for when the index does not exist

### DIFF
--- a/components/consumers/elasticsearch/main.go
+++ b/components/consumers/elasticsearch/main.go
@@ -10,12 +10,14 @@ import (
 	"strings"
 	"time"
 
+	elasticsearch "github.com/elastic/go-elasticsearch/v8"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
+	"github.com/go-errors/errors"
+
 	v1 "github.com/ocurity/dracon/api/proto/v1"
 	"github.com/ocurity/dracon/components/consumers"
 	"github.com/ocurity/dracon/pkg/enumtransformers"
 	"github.com/ocurity/dracon/pkg/templating"
-
-	elasticsearch "github.com/elastic/go-elasticsearch/v8"
 )
 
 var (
@@ -61,19 +63,21 @@ func main() {
 	flag.Parse()
 
 	if err := parseFlags(); err != nil {
-		log.Fatal(err)
+		log.Fatalf("could not parse incoming flags error: %s", err)
 	}
 
+	slog.Debug("connecting to elasticsearch")
 	es, err := getESClient()
 	if err != nil {
-		log.Fatal("could not contact remote Elasticsearch: ", err)
+		log.Fatalf("could not contact remote Elasticsearch error: %s", err)
 	}
+	slog.Debug("successfully connected to elasticsearch")
 
 	if consumers.Raw {
 		slog.Debug("Parsing Raw results")
 		responses, err := consumers.LoadToolResponse()
 		if err != nil {
-			log.Fatal("could not load raw results, file malformed: ", err)
+			log.Fatalf("could not load raw results, file malformed: %s", err)
 		}
 		numIssues := 0
 		for _, res := range responses {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/consumers/elasticsearch/main.go [hunk 1]
```diff
@@ -10,12 +10,15 @@ import (
...
 
+	"github.com/go-errors/errors"
+
 	v1 "github.com/ocurity/dracon/api/proto/v1"
 	"github.com/ocurity/dracon/components/consumers"
 	"github.com/ocurity/dracon/pkg/enumtransformers"
 	"github.com/ocurity/dracon/pkg/templating"
 
 	elasticsearch "github.com/elastic/go-elasticsearch/v8"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
```
ptzianos, line 21: nit: these imports should be grouped with the `go-errors` one.

(also, this is something that should be done automatically by the formatter but it's not being picked up)

components/consumers/elasticsearch/main.go
```diff
@@ -61,19 +64,24 @@ func main() {
...
+		slog.Error("could not parse incoming flags", slog.Any("error", err))
 		log.Fatal(err)
 	}
 
+	slog.Debug("connecting to elasticsearch")
 	es, err := getESClient()
 	if err != nil {
-		log.Fatal("could not contact remote Elasticsearch: ", err)
+		slog.Error("could not contact remote Elasticsearch: ", slog.Any("error", err))
+		log.Fatal(err)
```
ptzianos, line 75: do we need both the logging and the log.Fatal invocation? won't the `log.Fatal` just print the error anyway?

northdpole: great question and something i've been meaning to ask!
I want a better way to exit, ideally all of those would be cobra/viper without a log.Fatal but with a return err, therefore I think the structured logging here makes sense given a general plan to remove the "fatal" log line on 75

components/consumers/elasticsearch/main.go
```diff
@@ -61,19 +63,22 @@ func main() {
...
-		log.Fatal(err)
+		log.Fatalf("could not parse incoming flags error: %s", err)
 	}
 
+	slog.Debug("connecting to elasticsearch")
 	es, err := getESClient()
 	if err != nil {
-		log.Fatal("could not contact remote Elasticsearch: ", err)
+		log.Fatalf("could not contact remote Elasticsearch error: %s", err)
+		log.Fatal(err)
```
ptzianos, line 73: this is a duplicate

ptzianos (approved):

PR #395, opened 2024-10-03
393 fix producer base path

northdpole:


components/enrichers/custom-annotation/main_test.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, f := range files {
+		if strings.HasSuffix(f.Name(), ".raw.pb") {
+			continue
+		}
+		encodedProto, err := os.ReadFile(fmt.Sprintf("%s/%s", outdir, f.Name()))
+		require.NoError(t, err)
+
+		require.NoError(t, proto.Unmarshal(encodedProto, &actual))
+		if actual.OriginalResults.ToolName == expected[0].OriginalResults.ToolName {
+			if !proto.Equal(&actual, expected[0]) {
```
ptzianos, line 128: this could be simplified to `require.True(t, proto.Equal(&actual, expected[0]), cmp.Diff(&actual, expected[0], protocmp.Transform()))`

components/enrichers/custom-annotation/main_test.go
```diff
@@ -0,0 +1,139 @@
...
+	var actual draconv1.EnrichedLaunchToolResponse
+	for _, f := range files {
+		if strings.HasSuffix(f.Name(), ".raw.pb") {
+			continue
+		}
+		encodedProto, err := os.ReadFile(fmt.Sprintf("%s/%s", outdir, f.Name()))
+		require.NoError(t, err)
+
+		require.NoError(t, proto.Unmarshal(encodedProto, &actual))
+		if actual.OriginalResults.ToolName == expected[0].OriginalResults.ToolName {
```
ptzianos, line 127: you could use a map mapping the tool names to the expected results instead of using indices which are hard to follow

components/producers/producer_test.go
```diff
@@ -75,9 +75,9 @@ func TestWriteDraconOut(t *testing.T) {
 	require.NoError(t, proto.Unmarshal(pBytes, &res))
 
 	assert.Equal(t, "dracon-test", res.GetToolName())
-	assert.Equal(t, "./foobar", res.GetIssues()[0].GetTarget())
-	assert.Equal(t, "./barfoo", res.GetIssues()[0].GetTitle())
-	assert.Equal(t, "./example.yaml", res.GetIssues()[0].GetDescription())
+	assert.Equal(t, "foobar", res.GetIssues()[0].GetTarget())
```
andream16, line 78: Can we `require.NotEmpty(t, res.GetIssues())` to avoid potential panics from accessing `res.GetIssues()[0]`?

It would also help using an helper var to store the first issue :)

components/enrichers/custom-annotation/main.go
```diff
@@ -0,0 +1,79 @@
...
+	if annotation == "" {
+		annotation = defaultAnnotation
+	}
+	for _, r := range res {
+		slog.Info("processing results for ", slog.Any("scan", r.ScanInfo))
+		enrichedIssues := []*apiv1.EnrichedIssue{}
+		for _, i := range r.GetIssues() {
+			eI, err := enrichIssue(i)
+			if err != nil {
+				slog.Error(err.Error())
```
andream16, line 52: It would be nicer to add some extra details to this error so that we can identify it better

components/enrichers/custom-annotation/main.go
```diff
@@ -0,0 +1,79 @@
...
+	res, err := enrichers.LoadData()
+	if err != nil {
+		return err
+	}
+	if annotation == "" {
+		annotation = defaultAnnotation
+	}
+	for _, r := range res {
+		slog.Info("processing results for ", slog.Any("scan", r.ScanInfo))
+		enrichedIssues := []*apiv1.EnrichedIssue{}
```
andream16, line 48: ⛏️ since we know how many `res` we have we can init this more efficiently using `make([]*apiv1.EnrichedIssue, 0, len(res))`

components/enrichers/custom-annotation/main.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+			enrichedIssues = append(enrichedIssues, eI)
+		}
+
+		err := enrichers.WriteData(&apiv1.EnrichedLaunchToolResponse{
+			OriginalResults: r,
+			Issues:          enrichedIssues,
+		}, strings.ReplaceAll(annotation, " ", "-"))
+		if err != nil {
+			return err
```
andream16, line 63: Since we're in a loop here, should we log and continue?

components/enrichers/custom-annotation/task.yaml
```diff
@@ -0,0 +1,33 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: enricher-custom-annotation
+  labels:
+    v1.dracon.ocurity.com/component: enricher
+spec:
+  description: Identifies a code owner for each finding.
```
andream16, line 9: This needs to be updated 📚

components/enrichers/custom-annotation/task.yaml
```diff
@@ -0,0 +1,33 @@
...
+  params:
+  - name: enricher-custom-annotation-annotation
```
andream16, line 11: 🤔 this is a bit confusing to read, should we rename it to something easier to glance to?

Perhaps `enricher-custom-annotation-base-annotation`

components/enrichers/custom-annotation/README.md
```diff
@@ -0,0 +1,5 @@
+# Custom Annotation Enricher
+
+This enricher adds a custom annotation to every finding
+Useful for development purposes and for tagging results
+Can be used multiple times to add multiple annotations
```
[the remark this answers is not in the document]
northdpole: resolved, this now adds multiple annotations

components/enrichers/custom-annotation/main.go
northdpole: sorted f2f: if a component cannot write intermediate data the workflow is corrupted and should fail to be restarted

andream16 (approved):

PR #403, opened 2024-10-07
GitHub producer integration

northdpole:
test with running the workflow in `examples/github-security-project`, point the workflow to the `e2e-monorepo` branch `main`


components/producers/github-codeql/main.go
```diff
@@ -49,27 +63,36 @@ func main() {
...
 		if err != nil {
 			return nil, err
 		}
 
+		if resp.StatusCode > 299 {
+			body, bodyReadErr := io.ReadAll(resp.Body)
+			if bodyReadErr != nil {
+				return nil, errors.Errorf("could not read response error from github, err:%w", bodyReadErr)
+			}
+			return nil, errors.Errorf("did not receive a valid status code from github, status code: %d, body:%s", resp.StatusCode, body)
```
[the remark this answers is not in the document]
northdpole: usually, it's just some json with a short but helpful error message, i prefer it in the error since the error is our logs but i can also add a log line

pkg/utils/test.go
```diff
@@ -0,0 +1,15 @@
+package utils
+
+import "time"
+
+func PointerFromFloat64(a float64) *float64 {
+	return &a
+}
+
+func TimeMustParse(layout, timestamp string) time.Time {
```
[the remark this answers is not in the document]
northdpole: was trying to be clever and create a helper for this sort of thing but happy to remove

components/producers/github-codeql/main.go
```diff
@@ -49,27 +65,40 @@ func main() {
...
+		ctx, cancel := context.WithTimeout(ctx, time.Second*30)
+		defer cancel()
+
+		alerts, resp, err := apiClient.ListRepoAlerts(ctx, owner, repo, opt)
 		if err != nil {
 			return nil, err
 		}
 
+		if resp.StatusCode > 299 {
+			body, bodyReadErr := io.ReadAll(resp.Body)
```
andream16, line 94: We should close the body to avoid leaks

`defer resp.body.Close()`

components/producers/github-codeql/main.go
```diff
@@ -24,17 +27,30 @@ var (
...
 	flag.StringVar(&GitHubToken, "github-token", "", "The GitHub token used to authenticate with the API")
+	flag.StringVar(&Ref, "reference", "", "The Ref/branch to get alerts for")
+	flag.StringVar(&Severity, "severity", "", "If specified, only code scanning alerts with this severity will be returned. Possible values are: critical, high, medium, low, warning, note, error")
+	toolName = "CodeQL"
 	if err := producers.ParseFlags(); err != nil {
 		log.Fatal(err)
 	}
-
-	alerts, err := listAlertsForRepo(RepositoryOwner, RepositoryName, GitHubToken)
+	ctx := context.WithoutCancel(context.Background())
```
andream16, line 51: You should use a `context.WithCancel()` here and `defer cancel()`

This will be only called when your app exists so it's all good but in a different scenario you can basically called `cancel` whenever you want and signal a bunch of listening components to shutdown and exit gracefully

components/producers/github-codeql/main.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+		slog.Error("github response", slog.String("body", string(body)))
+		return nil, nil, errors.Errorf("did not receive a valid status code from github, status code: %d, body:%s", resp.StatusCode, string(body))
+	}
+	return alerts, resp, nil
+}
+
+func listAlertsForRepo(ctx context.Context, apiClient wrapper.Wrapper, owner, repo, toolName, severity string) ([]*github.Alert, error) {
+	var allAlerts []*github.Alert
+	ctx, cancel := context.WithTimeout(ctx, time.Minute*5)
```
andream16, line 91: It would be nice to make this configurable via env var `GITHUB_CLIENT_REQUEST_TIMEOUT=5m`.

It would need to be parsed like https://github.com/ndisidore/go-env/blob/main/parser.go#L71

components/producers/github-codeql/main.go
```diff
@@ -0,0 +1,179 @@
...
+	defer cancel()
+
+	opt := &github.AlertListOptions{
+		State:    "open",
+		Ref:      cfg.Ref,
+		Severity: severity,
+		ToolName: toolName,
+		ListOptions: github.ListOptions{
+			PerPage: 100,
```
andream16, line 100: Would be cool to configure this by env var too i.e. `GITHUB_CLIENT_LIST_PAGE_SIZE=100`

components/producers/github-codeql/main.go
```diff
@@ -0,0 +1,179 @@
...
+	default:
+		return v1proto.Severity_SEVERITY_UNSPECIFIED
+	}
+}
+
+func parseGithubCWEsFromTags(tags []string) []int32 {
+	// example input: ["security", "external/cwe/cwe-022"]
+	cwePrefix := "external/cwe/cwe-"
+
+	cwes := []int32{}
```
andream16, line 165: ⛏️ this could be `var cews = make([]int32, 0, len(tags))`

components/producers/github-codeql/main.go
```diff
@@ -0,0 +1,179 @@
...
+	for _, item := range tags {
+		if strings.HasPrefix(item, cwePrefix) {
+			cweString := strings.TrimPrefix(item, cwePrefix)
+
+			cwe, err := strconv.Atoi(cweString)
+			if err != nil {
+				slog.Warn("Failed to parse CWE from tag", "tag", item, "error", err)
```
andream16, line 172: We can use the following for consistency and better fields parsing

```
slog.Warn("Failed to parse CWE from tag", slog.String("tag", item), slog.Any("error", err))
```

components/producers/github-dependabot/main.go
```diff
@@ -0,0 +1,169 @@
...
+	opt := &github.ListAlertsOptions{
+		State:     &open,
+		Severity:  &severity,
+		Ecosystem: &ecosystem,
+		ListOptions: github.ListOptions{
+			PerPage: 30,
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, time.Minute*5)
```
andream16, line 80: Same comment here

components/producers/github-dependabot/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+func parseIssues(alerts []*github.DependabotAlert) []*v1proto.Issue {
+	issues := make([]*v1proto.Issue, 0, len(alerts))
+	for _, alert := range alerts {
+		ecosystem := *(alert.GetSecurityVulnerability().Package.Ecosystem)
+		if ecosystem == "pip" {
+			ecosystem = "pypi"
+		}
+
+		cwe := []int32{}
```
andream16, line 110: Same comment as above

components/producers/github-dependabot/main.go
```diff
@@ -0,0 +1,169 @@
...
+			if err != nil {
+				slog.Error("could not extract cwe number from ", slog.String("cweID", *c.CWEID))
+				continue
+			}
+
+			cwe = append(cwe, int32(cweNum))
+		}
+		purl := producers.GetPURLTarget(ecosystem, "", *alert.GetSecurityVulnerability().Package.Name, "", packageurl.Qualifiers{}, "")
+		slog.Info("Handling alert for", slog.String("purl", purl))
+		cve := ""
```
andream16, line 123: super ⛏️  but we could use a `var ()` block to define all these vars inline

Would look slicker imo

```
var (
    cve string
    summary string 
    ...
)
```

components/producers/github-dependabot/main.go
```diff
@@ -0,0 +1,169 @@
...
+		if advisory != nil {
+			if advisory.CVEID != nil {
+				cve = *advisory.CVEID
+			} else if advisory.GHSAID != nil {
+				cve = *advisory.GHSAID
+			}
+
+			description = *advisory.Description
+			summary = *advisory.Summary
+			cvss = *advisory.CVSS.Score
```
andream16, line 138: Can CVSS be nil? If yes we should add a check to avoid panics

andream16 (approved):

PR #404, opened 2024-10-08
[DRAFT] - Adding Go Component SDK proposal

andream16:
Documenting the proposal for the Go Component SDK.

Please, read the Objectives and the Callouts before reviewing deeply. This is important to set expectations and move the bar a bit higher without trying to design too much too fast.

The objective is to develop the proposal incrementally and learn more about the final SDK while migrating our components over.

I expect the final result to be a bit different from the proposal but still follow the same direction.

---

For any not clear bit feel free to leave a comment or send me a message in chat :)

---

I think it's better if you review it by glancing at the actual rendered Markdown :)

### DIFF
--- a/docs/dep/dep-006-new-go-sdk.md
+++ b/docs/dep/dep-006-new-go-sdk.md
@@ -1,0 +1,747 @@
+# New Go SDK Design
+
+This document highlights the work to be done to create a new Go SDK to write Dracon components.
+
+**Table of contents:**
+
+* [Objectives](#objectives)
+* [Callouts](#callouts)
+* [Proposal](#proposal)
+  * [Structure and contributions guidelines](#proposal-structure)
+  * [Component API](#proposal-component-api)
+    * [First pass - simple components](#proposal-component-api-first-pass)
+      * [Creating a new component](#proposal-component-api-first-pass-create)
+        * [Running a component](#proposal-component-api-first-pass-running)
+        * [Defaults](#proposal-component-api-first-pass-defaults)
+        * [Example](#proposal-component-api-first-pass-example)
+    * [Second pass - environment aware components](#proposal-component-api-second-pass)
+      * [Why the first pass is not enough](#proposal-component-api-second-pass-why)
+      * [A new environment aware runner](#proposal-component-api-second-pass-env-aware-runner)
+      * [Creating a new component](#proposal-component-api-second-pass-create-new-component)
+        * [Example](#proposal-component-api-second-pass-example)
+        * [Wrapping it up](#proposal-component-api-second-pass-wrap)
+  * [Components catalog](#proposal-components-catalog)
+  * [Default readers/writers](#proposal-default-readers-writers)
+    * [Local](#proposal-default-readers-writers-local)
+    * [SAAS](#proposal-default-readers-writers-saas)
+
+<a id="objectives"></a>
+
+## Objectives
+
+* Create an SDK to write components in a predictable way
+* The SDK should take care of how a component runs and should handle:
+  * logging, monitoring and tracing
+  * panics
+  * graceful shutdown and cancellations
+  * health checks
+* Provide a set of reusable building blocks to perform common operations, such as reading from a source
+* Provision a simple default storage (overridable) to allow transferring information between different components
+* Standardise environment aware components: initialisation, execution and shutdown
+* Represent information shared between components using [ocsf](https://github.com/ocsf/examples/tree/main) format
+  leveraging protobuf and generated code
+* Make SDK and components easy to test
+* Work towards being able to run components with zero dependencies locally (single binary)
+
+<a id="callouts"></a>
+
+## Callouts
+
+* This document covers ONLY how to create new components via SDK
+* The advertised proposal will be worked on incrementally
+* Some more technical bits about retries, metrics and so on have been left out for the moment.
+  We'll discuss them as a follow-up in RFCs.
+* The implementation will likely differ a bit from the proposal as we'll tweak the final the SDK
+  while migrating components. We'll track the various changes to the SDK with RFCs.
+* The proposed environment variables will be heavily documented and validated
+* A default `context.Context` is being currently passed around in the examples for the SDK.
+  This will most likely change with an internal `component.Context` which still preserves `context.Context`
+  semantics like cancellation etc but also provides stronger typing for context values which will be
+  most likely used to get re-usable debugging/monitoring bits around the components. I.E. `component.GetLoggerFromContext(ctx)`.
+  This is similar to how lots of workflow engines execute on this, for example [Temporal](https://typescript.temporal.io/api/classes/activity.Context).
+
+<a id="proposal"></a>
+
+## Proposal
+
+<a id="proposal-structure"></a>
+
+### Structure and contributions guidelines
+
+The code for the new SDK will live in this repository in the `sdk/component` directory.
+
+The SDK will have its own go module `github.com/ocurity/dracon/sdk/component` with its own set of dependencies.
+
+> Why?
+>
+> * having a separate directory helps keeping responsibilities segregated
+> * having a separate module helps keeping dependencies separate, we want very few dependencies on the core SDK to boost adoption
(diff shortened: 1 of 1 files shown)


docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+	"github.com/ocurity/dracon/sdk/component"
+	"github.com/ocurity/dracon/api/gen/com/github/ocsf/ocsf-schema/v1"
+)
+
+// simpleComponent parses a Vulnerability Findings report from JSON
+// enriches it and stores the processed output to SQLite.
+type simpleComponent struct{}
+
+func (s simpleComponent) Read(ctx context.Context) (*v1.VulnerabilityFinding, error) {
+	b, err := os.ReadFile("input.json")
```
northdpole, line 327: Mirroring offline conversation:
Can we please abstract the read to the sdk? with this implementation, if we ever change where we read things from we need to rewrite every component to match, if Read just has to call a "Give me data" we can then easily abstract where we read data from to the sdk and make it a configuration details

docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+		return nil, fmt.Errorf("failed to read input: %w", err)
+	}
+
+	// Parse to *v1.VulnerabilityFinding
+
+	return &v1.VulnerabilityFinding{}, nil
+}
+
+func (s simpleComponent) Write(ctx context.Context, out *v1.VulnerabilityFinding) error {
+	// Write finding to SQLite
```
northdpole, line 338: same as above for write

docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+│       ├── tagger/
+│       │   └── go.mod
+│       └── prioritiser/
+│           └── go.mod
+```
+
+The components will follow the known structure and just re-use actions or define their own
+to implement the goals of the component.
+
+One can see a scenario where an SQLite reader/writer can be used between multiple components
```
northdpole, line 420: great idea! i'd suggest we make where we read/write an implementation detail hidden by the SDK, then the components do not need to implement multiple read/writers and instead can focus on deep transformation and integration with the target things

docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+
+### Component API
+
+At its core, regardless of the type of the component, each one of them has to perform 3 actions:
+
+* `Read`: read raw information from some source - filesystem, database, api, ...
+* `Transform`: perform operations to enrich the raw data based on the context - adding annotations, call an api, ...
+* `Write`: write the enriched results - filesystem, database, api ...
+
+Different components implement these actions in one way or another. Some of them can be NOOP.
```
dlicheva, line 103: In what case would the components be NOOP?

docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+
+The components share information between different actions. This information follows a predictable format of a vulnerability report.
+
+This means that we can provide a minimum set of API to guide the creation of new components. The Component API will be represented by a `Runner` interface which can be satisfied to implement a new component:
+
+```go
+// Runner advertises the actions that need to be implemented by a Smithy component.
+type Runner interface {
+    // Read should be implemented to read Smithy's raw input from an underlying data source.
+    Read(ctx context.Context) (*v1.VulnerabilityFinding, error)
```
dlicheva, line 113: The runner looks great for vulnerability scanners, but some scanners return other data, like SBOMs. Would they need to implement a different runner? Or should it be something like "DataFinding"?

docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+* [setting GOMAXPROCS to match Linux container CPU quota](https://github.com/uber-go/automaxprocs)
+* exposing/pushing metrics + registering default metrics
+* exposing health checks for long-lived components
+* emitting traces
+* profiling server
+* ...
+* supporting any other reliability capability
+
+```go
+func (c *component) Run(ctx context.Context, runner Runner) error {
```
dlicheva, line 201: This is a beautiful example :star:

docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+│   │   │   └── go.mod
+│   │   └── elasticsearch/
+│   │       └── go.mod
+│   └── transformer/
+│       ├── atom-enricher/
+│       │   └── go.mod
+│       ├── tagger/
+│       │   └── go.mod
+│       └── prioritiser/
+│           └── go.mod
```
dlicheva, line 414: This looks good, but my only concern here is that if I want to find all parts of a component, I need to dig around for them in different folders. What do you think about the alternative?:
```
├── sqlite/
│     ├── reader.mod
│     ├── transformer.mod
│     └── writer.mod
```

docs/dep/dep-006-new-go-sdk.md
```diff
@@ -0,0 +1,474 @@
...
+
+It's great in environments where a database service is not necessary.
+
+*We need to do a spike with SQLite first to finalise the decision or pivot fast.*
+
+<a id="proposal-default-readers-writers-saas"></a>
+
+#### SAAS
+
+For `SAAS` we decided to invest in a more sophisticated system to abstract reads/writes to different places.
```
dlicheva, line 464: Would the SAAS include an SQLite db, to fall back on until the custom ones are configured? Thinking about what will make the SAAS easier to onboard.

dlicheva (commented):
Thank you Andrea, the proposal looks good! I've just left a couple of questions.

docs/dep/dep-006-new-go-sdk.md
andream16: My understanding is that right now we want to get started with [VulnerabilityFinding](https://schema.ocsf.io/1.3.0/classes/vulnerability_finding) from [oscf](https://schema.ocsf.io/) and this can be used to represent all the inflight data that we need between components (cc @northdpole )

In any case, if we wanted to cover other categories, we could easily do so by leveraging an enum

I think for now this is good enough but let me know what you think!

docs/dep/dep-006-new-go-sdk.md
andream16: Yes, will extend the doc with this section :)

docs/dep/dep-006-new-go-sdk.md
andream16: Makes sense! It would still be nice to separate modules for cases where, for example, a reader has a lot of dependencies and the writer has way less - I think for now anything works!

docs/dep/dep-006-new-go-sdk.md
andream16: I think that to make SAAS work asap we'd have to write an integration for Postgres, to comply to how it works today - @ptzianos correct me if I'm wrong

SQLite is not designed to be used reliably by multiple applications - it's good enough for a standalone app that is "ok" losing data

docs/dep/dep-006-new-go-sdk.md
andream16: Done

docs/dep/dep-006-new-go-sdk.md
andream16: Updated

docs/dep/dep-006-new-go-sdk.md
northdpole: I agree with @andream16 , for now, VulnFinding is good enough, for communicating sboms, since ocsf doesn't give you any way to attach a cyclonedx doc, I think we can use the ["observables"](https://schema.ocsf.io/1.3.0/objects/observable?extensions=) array to communicate a serialized Cyclonedx document

ptzianos (changes requested):

PR #410, opened 2024-10-10
close #409, refactor Jira consumer,

### LINKED ISSUE #409: Refactor Jira Consumer, improve templating and UX
the jira consumer is pretty unwieldy, the config is a json string which does not make much sense and it has both a "description template" and a "add to description" part which conflict and are not clear on what they do.<br><br>Instead:

* add a pretty ok default description template
* remove the "add to description functionality

northdpole:
remove addToDescription, instead offer a default template
now the only piece of jira config needed to run the consumer is the project name

### DIFF
--- a/pkg/jira/jira/apiutils.go
+++ b/pkg/jira/jira/apiutils.go
@@ -2,8 +2,8 @@ package jira
 
 import (
 	"bytes"
+	_ "embed"
 	"encoding/json"
-	"fmt"
 	"log"
 	"path/filepath"
 	"strconv"
@@ -29,6 +29,9 @@ type defaultJiraFields struct {
 	Labels          []string
 	CustomFields    tcontainer.MarshalMap
 }
+
+//go:embed issueTemplate.txt
+var defaultTemplate string
 
 // getDefaultFields creates the fields for Project, IssueType, Components, AffectsVersions, Labels and CustomFields
 // with the default values specified in config.yaml and serializes them into Jira Fields.
@@ -112,7 +115,11 @@ func draconResultToSTRMaps(draconResult document.Document) (map[string]string, s
 }
 
 // makeDescription creates the description of an issue's enhanced with extra information from the Dracon Result.
-func makeDescription(draconResult document.Document, extras []string, template string) string {
+func makeDescription(draconResult document.Document, template string) string {
+
+	if template == "" {
+		template = defaultTemplate
+	}
 	if draconResult.Count == "" {
 		draconResult.Count = "0"
 	}
@@ -131,6 +138,7 @@ func makeDescription(draconResult document.Document, extras []string, template s
 	if err != nil {
 		log.Fatal("could not template enriched issue ", err)
 	}
+
 	description, err := templating.TemplateStringEnriched(template,
 		&v1.EnrichedIssue{
 			Annotations:   draconResult.Annotations,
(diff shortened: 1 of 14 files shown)


pkg/jira/jira/apiutils.go
```diff
@@ -112,7 +113,25 @@ func draconResultToSTRMaps(draconResult document.Document) (map[string]string, s
 }
 
 // makeDescription creates the description of an issue's enhanced with extra information from the Dracon Result.
-func makeDescription(draconResult document.Document, extras []string, template string) string {
+func makeDescription(draconResult document.Document, template string) string {
```
andream16, line 116: f2f exploring go embed

pkg/jira/jira/issueTemplate.txt
```diff
@@ -0,0 +1,8 @@
+{{.ToolName}} detected '{{.RawIssue.Title}}' at {{.RawIssue.Target}} during scan started at: {{.ScanStartTime}} with id {{.ScanID}}.
```
andream16, line 1: ❤️

pkg/templating/template_description.go
```diff
@@ -33,8 +36,98 @@ func TemplateStringRaw(inputTemplate string, issue *v1.Issue) (*string, error) {
...
+	return func(ei *enrichedIssue) error {
+		if toolname == "" {
+			return errors.New("invalid toolname ''")
+		}
+		ei.ToolName = toolname
+		return nil
+	}
+}
+
+func EnrichedIssueWithScanStartTime(startTime *time.Time) enrichedIssueOption {
```
andream16, line 61: Perhaps this should be a pointer? 

It would be not super ergonomic to use from an API POV. Imagine having to pass a reference to a `time.Date`

I'd keep a value and check that it's not a 0 value :)

pkg/templating/template_description.go
```diff
@@ -33,8 +36,98 @@ func TemplateStringRaw(inputTemplate string, issue *v1.Issue) (*string, error) {
...
+	return func(ei *enrichedIssue) error {
+		if severity == "" {
+			return errors.New("invalid severity ''")
+		}
+		ei.SeverityText = severity
+		return nil
+	}
+}
+
+func EnrichedIssueWithCount(count int) enrichedIssueOption {
```
andream16, line 91: ⛏️ We can use an uint here to make it obvious that this has to be 0 or positive

pkg/templating/template_description.go
```diff
@@ -33,8 +36,98 @@ func TemplateStringRaw(inputTemplate string, issue *v1.Issue) (*string, error) {
...
+	return func(ei *enrichedIssue) error {
+		if scanID == "" {
+			return errors.New("invalid scanID ")
+		}
+		ei.ScanID = scanID
+		return nil
+	}
+}
+
+func EnrichedIssueWithFirstFound(firstFound *time.Time) enrichedIssueOption {
```
andream16, line 111: Same comment here as on the other about time

pkg/templating/template_description.go
```diff
@@ -33,8 +36,98 @@ func TemplateStringRaw(inputTemplate string, issue *v1.Issue) (*string, error) {
...
+
 // TemplateStringEnriched applies the provided go template to the Enriched Issue provided and returns the resulting str
-func TemplateStringEnriched(inputTemplate string, issue *v1.EnrichedIssue) (*string, error) {
+func TemplateStringEnriched(inputTemplate string, issue *v1.EnrichedIssue, opts ...enrichedIssueOption) (*string, error) {
+	enrichedIssue := &enrichedIssue{
+		EnrichedIssue: issue,
+	}
+	for _, opt := range opts {
+		if err := opt(enrichedIssue); err != nil {
+			return nil, err
```
andream16, line 128: I'd use something like `return nil, fmt.Errorf("could not apply enriched issue option: %w", err)` to make it more obvious where the error is coming from

andream16 (approved):

PR #423, opened 2024-10-15
fix #422, allow consumer templating to also carry information about w…

### LINKED ISSUE #422: consumer-templating should also allow for defining which annotation carries the fix for the vulnerability

northdpole:
…hich annotation contains the code fix

### DIFF
--- a/pkg/templating/template_description.go
+++ b/pkg/templating/template_description.go
@@ -20,13 +20,14 @@ const (
 type (
 	enrichedIssue struct {
 		*v1.EnrichedIssue
-		ToolName       string
-		ScanStartTime  string
-		ScanID         string
-		ConfidenceText string
-		SeverityText   string
-		Count          uint
-		FirstFound     string
+		ToolName            string
+		ScanStartTime       string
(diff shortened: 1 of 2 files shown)


pkg/templating/template_description.go
```diff
@@ -126,6 +127,14 @@ func EnrichedIssueWithFirstFound(firstFound time.Time) enrichedIssueOption {
 	}
 }
 
+// EnrichedIssueWithFixedCodeAnnotation allows to define which of the annotations contains fixed code
+func EnrichedIssueWithFixedCodeAnnotation(fixedCodeAnnotation string) enrichedIssueOption {
+	return func(ei *enrichedIssue) error {
+		ei.FixedCodeAnnotation = fixedCodeAnnotation
```
andream16, line 133: Should we check for the string not to be empty?

northdpole: this annotation could easily be empty however, not every workflow needs to have a component that generates fixed code :-)

andream16 (approved):

PR #424, opened 2024-10-15
Generating OCSF Go types from Proto Schema.

andream16:
Preparing for OSS work by generating Go types from [OSCF proto schema definition](https://github.com/ocsf/examples/tree/main/encodings/protobuf/proto).

Note: 
`buf` supports [git](https://buf.build//docs/reference/inputs#git) sources but funnily enough doesn't allow us to specify a destination for the generated code (like it usually happens for local development). Since buf relies on git for this remote generation, I added a bunch of stuff to the scratch image to allow so (I'm open to just use the base alpine image to simplify things if you think that it's tricky to understand).

---

>Why remote generation and not just copying over the type?
In this way we don't need to sync up manually and we can also target specific versions.

>Is there a buf schema registry for ocsf?
Nope https://buf.build/search?query=ocsf

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -6,3 +6,7 @@ plugins:
 inputs:
   - proto_file: api/proto/v1/issue.proto
   - proto_file: api/proto/v1/engine.proto
+  - git_repo: https://github.com/ocsf/examples
+    branch: main
+    ref: ba2a49f2bb1faf0c75ece9b6a9c5cd608eafbc67
+    subdir: encodings/protobuf/proto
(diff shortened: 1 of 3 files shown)


buf.gen.yaml
```diff
@@ -6,3 +6,7 @@ plugins:
 inputs:
   - proto_file: api/proto/v1/issue.proto
   - proto_file: api/proto/v1/engine.proto
+  - git_repo: https://github.com/ocsf/examples
+    branch: main
+    ref: ba2a49f2bb1faf0c75ece9b6a9c5cd608eafbc67
```
andream16, line 11: They don't use tags but we can still pin a commit so we have some sort of versioning.

ptzianos (approved):

PR #425, opened 2024-10-16
OCU-219 - sdk interfaces

andream16:
- Moving OCSF generated code to the sdk to keep them used where needed.
- Defined interfaces for all components.
- Renamed components based on team discussion to keep names and implementations closer to their actually behaviour.

---

This approach comes after discussing with the team and @ptzianos and acknowledges feedback from https://github.com/ocurity/dracon/pull/404/files

The approach enables to:
- define clearer behaviours for components
- keep dependencies sorted: the main sdk will not know anything about actual underlying component dependencies

---

I'll leave some open questions as comments, please review them!

---

I'll cleanup the commits before merging.


sdk/component/go.mod
```diff
@@ -0,0 +1,14 @@
+module github.com/ocurity/dracon/sdk/component
```
andream16, line 1: ❓  In my head it makes sense to have a `sdk/component` module but I also can see a world where we just have a `sdk` module, in case we want to add more functionalities to the SDK (for example for workflow management).

I still think that the first option makes more sense as it's more isolated. Let me know what you think.

sdk/component/component.go
```diff
@@ -0,0 +1,71 @@
...
+		Filter(ctx context.Context, findings []*ocsf.VulnerabilityFinding) ([]*ocsf.VulnerabilityFinding, error)
+	}
+)
+
+// Components interfaces.
+type (
+	// Sourcer defines the behaviour of source components.
+	Sourcer interface {
+		// Source runs an arbitrary step. Useful for interacting with third-party API - i.e. cloning a repository.
+		Source(ctx context.Context) error
```
andream16, line 35: ❓ I'm not a super fan of this method name so please come up with better suggestions if you think of any

sdk/component/component.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// Filterer allows filtering out findings by a specified criteria.
+	Filterer interface {
+		// Filter returns filtered findings from the supplied ones.
+		Filter(ctx context.Context, findings []*ocsf.VulnerabilityFinding) ([]*ocsf.VulnerabilityFinding, error)
+	}
+)
+
+// Components interfaces.
+type (
```
andream16, line 31: ❓ Should we just merge the concepts of enrichers and producers?

I think that they do very similar things and they only differ in semantics:
- a producer writes vs an enricher updates
- an enricher may filter vs a producer doesn't

Maybe we could unify them under a `Processor` interface and compose the behaviours.

sdk/component/component.go
```diff
@@ -0,0 +1,71 @@
...
+	Sourcer interface {
+		// Source runs an arbitrary step. Useful for interacting with third-party API - i.e. cloning a repository.
+		Source(ctx context.Context) error
+	}
+
+	// Producer defines the behaviour of producer components.
+	// Read -> Process -> Validate -> Store.
+	Producer interface {
+		Reader
+		Validator
```
andream16, line 42: ❓ I think it makes sense to only force people to validate on write as otherwise we'd have to have steps for validating inputs and outputs on `Producer`, `Consumer` and `Enricher`. I think that for now this is enough.

sdk/component/component.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	// Consumer defines the behaviour of consumer components.
+	// Read -> Process.
+	Consumer interface {
+		Reader
+
+		// Process processes the findings and takes some action. Useful for interacting with third-party API or
+		// creating results - i.e. raising tickets.
+		Process(ctx context.Context, findings []*ocsf.VulnerabilityFinding) error
```
andream16, line 69: ❓ I don't have a strong preference for this name. We could just rename it to `Consume`. Let me know what you think.

sdk/component/component.go
dlicheva: As discussed, I think we should validate only the input data of the producers, to check if we're feeding them meaningful values. The output shouldn't be validated, because we can't maintain what 3rd party tools produce.

sdk/component/component.go
dlicheva: What do you  think about "Write"?

sdk/component/go.mod
dlicheva: `sdk/component` is clearer IMHO.

andream16: Opted for module at `sdk` level so we can use the same module for a future `workflow` package and similar.

Still would work nice with `component.Stuff` or `workflow.Stuff` with imports

`github.com/ocurity/dracon/sdk/component` and github.com/ocurity/dracon/sdk/workflow

sdk/component/component.go
northdpole: a producer takes info from the outside and normalizes it
an enricher takes info from the inside and updates it with extra info

if we merge them, how are we going to differentiate what runs when?

sdk/component/component.go
northdpole: `ProcessSource` ?
A git-sourcer clones a repo and filters paths
A wget sourcers downloads results from a path
A potential AWS sourcer sets up access to aws
A potential X sourcer ensures that X is ready for scanning

sdk/component/component.go
northdpole: output?
Isn't a producers output Smithy-OCSF?

sdk/component/component.go
andream16: The producer still reads the parsed vuln findings, does something with them and outputs them somewhere (the enriched does a similar thing).

Could the different behaviours between them be just implementation details rather than being different components at all?

sdk/component/component.go
andream16: @dlicheva that's fair but how do we defend ourselves from having flows where we could write malformed data in the database?

@northdpole yes - but I think it would be better for data integrity to validate what we're about to output in the data destination - thoughts?

sdk/component/component.go
andream16: Not sure if a Consumer would write in most cases 🤔 

In my head a consumer does something with the available results, without any extra massage steps

For example: interprets the findings and opens a JIRA ticket for each one of them

andream16:
Opened https://github.com/ocurity/dracon/pull/435 to replace this

PR #435, opened 2024-10-18
OCU-219 - sdk interfaces

andream16:
- Moving OCSF generated code to the sdk to keep them used where needed.
- Defined interfaces for all components.
- Renamed components based on team discussion to keep names and implementations closer to their actually behaviour.

---

This approach comes after discussing with the team and @ptzianos and acknowledges feedback from https://github.com/ocurity/dracon/pull/404/files

The approach enables to:
- define clearer behaviours for components
- keep dependencies sorted: the main sdk will not know anything about actual underlying component dependencies


northdpole (approved):

sdk/component/component.go
```diff
@@ -0,0 +1,87 @@
...
+		Transform(ctx context.Context, payload Unmarshaler) (*ocsf.Vulnerability, error)
+	}
+
+	// Filter allows filtering out vulnerability findings by some criteria.
+	Filter interface {
+		Reader
+		Updater
+
+		// Filter returns filtered findings from the supplied ones applying some criteria.
+		// It returns false if no findings have been filtered out.
```
dlicheva, line 66: What's the difference in it returning false vs. returning an empty []?

dlicheva (approved):
It looks good and well structured :) I've just added one question from how I'm used to things in the JS world

sdk/component/component.go
andream16: I think that it would be nice to return `false` if nothing was filtered out and `true` if something was filtered out

I've seen this often in other packages as an approach and it simplifies utilisation imo :)

ptzianos (approved):

PR #441, opened 2024-10-22
Feature/440/changelog utility

### LINKED ISSUE #440: create changelog utility
would be nice to be able to autopublish our changelog to do this we need

* a way to autogenerate an editable changelog from commit messages
* a way to audit commit messages (and thus changelog)
* a way to publish our changelog with the latest release artifacts

northdpole:
close #440


andream16:
Can you post a sample output?

northdpole:
> Can you post a sample output?

absolutely, here you go
<details>

```
## v0.1.2.3
**example changelog**

* changelog binary that allows to generate a changelog consisting of commit messages between current HEAD and the last tag

* add changelog target to the makefile

* Update reviewdog references to smithy.

* Updating .gitignore to ignore smithy files previously ignored by ocurity.

* Updating actions to leverage smithy references.

* Updating tests to leverage smithy.

* Updating lock.

* Update README to replace ocurity with smithy.

* Updating protobuf contracts to use smithy as reference.

* Updating docs to use smithy references.

* Updating examples to leverage smithy.
```
</details>

andream16:
> > Can you post a sample output?
> 
> absolutely, here you go
> 
> `$changelog --repo-path . --new-tag-name="v0.1.2.3" --new-tag-message="example changelog"`
> 
> ```
> ## v0.1.2.3
> **example changelog**
> 
> * changelog binary that allows to generate a changelog consisting of commit messages between current HEAD and the last tag
> 
> * add changelog target to the makefile
> 
> * Update reviewdog references to smithy.
> 
> * Updating .gitignore to ignore smithy files previously ignored by ocurity.
> 
> * Updating actions to leverage smithy references.
> 
> * Updating tests to leverage smithy.
> 
> * Updating lock.
> 
> * Update README to replace ocurity with smithy.
> 
> * Updating protobuf contracts to use smithy as reference.
> 
> * Updating docs to use smithy references.
> 
> * Updating examples to leverage smithy.
> ```

I think this looks good but I'd like to not have blank lines between each commit report. It would also be cool to add a short timestamp to each commit, which can be helpful for debugging potential individual issues.

ptzianos:
As far as I understand the logic, the workflow is that the user runs the utility with the new tag that they want to add and the script generates a list of commit messages since the latest tag (latest being the latest one based on the timestamps)

What happens if we release a new major version (v1.0.0) but until it's rolled out everywhere we release a patch version of the previous major tag (previous latest version v0.66.0 and we release v0.66.1). This utility would produce list of commit messages for the new tag v1.0.1 from v0.66.1 to now instead of from v1.0.0, right? I would expect it to use the current tag passed as a flag to figure out that closest one, and the closest one would be found based on just simple string sorting. 

Also, I would expect that we could pass flags to just create a new major/minor/patch version based on the latest tag because that's the most common use case and it would allow us to remove the relevant make targets

northdpole:
> > > Can you post a sample output?
> > 
> > 
> > absolutely, here you go
> > `$changelog --repo-path . --new-tag-name="v0.1.2.3" --new-tag-message="example changelog"`
> > ```
> > ## v0.1.2.3
> > **example changelog**
> > 
> > * changelog binary that allows to generate a changelog consisting of commit messages between current HEAD and the last tag
> > 
> > * add changelog target to the makefile
> > 
> > * Update reviewdog references to smithy.
> > 
> > * Updating .gitignore to ignore smithy files previously ignored by ocurity.
> > 
> > * Updating actions to leverage smithy references.
> > 
> > * Updating tests to leverage smithy.
> > 
> > * Updating lock.
> > 
> > * Update README to replace ocurity with smithy.
> > 
> > * Updating protobuf contracts to use smithy as reference.
> > 
> > * Updating docs to use smithy references.
> > 
> > * Updating examples to leverage smithy.
> > ```
> 
> I think this looks good but I'd like to not have blank lines between each commit report.
> 
> It would also be cool to add a short timestamp to each commit, which can be helpful for debugging potential individual issues!
> 
> @northdpole

@andream16 , how about this? 
```
## v0.1.2.3

**foobar**

2024-10-22T12:20:21+01:00: changelog binary that allows to generate a changelog consisting of commit messages between current 
HEAD and the last tag
2024-10-22T12:18:06+01:00: add changelog target to the makefile
2024-10-31T11:02:12Z: workaround defectdojo expecting a non-empty scanstarttime for their engagments
```

docs/CHANGELOG.md
```diff
@@ -0,0 +1,332 @@
+## 
+
+**v0.57.0**
+2024-10-22T12:20:21+01:00: add changelog binary that allows to generate a changelog consisting of commit messages between defined tag and HEAD
```
andream16, line 4: Can we please use just `day-month-year` format? I think that we don't need the hours, minutes and seconds personally :)

docs/CHANGELOG.md
```diff
@@ -0,0 +1,332 @@
...
+2024-10-22T12:18:06+01:00: add changelog target to the makefile
+2024-10-16T11:33:41+01:00: remove results output from git-clone component
+**v0.56.5**
```
andream16, line 7: I'd add empty lines to separate the various releases

PR #443, opened 2024-10-23
Register and run components

andream16:
**PLEASE TAKE A LOOK AT THE README TO GET MORE CONTEXT BEFORE REVIEWING**

----

Closes OCU-251, OCU-254, OCU-220, OCU-230, OCU-232

---

**What's not covered**:
- Storage logic to actually persist vulnerability findings in the database (PSQL/SQLite)
- Metrics, Tracing handling

---

**What's covered**:
- Run components using the same run bed to configure common utilities and setup guardrails. Inject common log fields in the runners to have smart defaults useful for debugging and standardisation.
- Logging
- Panic handling
- Context cancellation and signal termination handling
- Graceful shutdown
- Auto initialisation of the runner based on defaults + environment variables and extra customisation via functional options.

---

**Technical choices**:
- Used [gomock](https://github.com/uber-go/mock) for mocking to boost productivity and not maintaining mocks manually. With gomock we can generate the mocks based on mockable interfaces. The way that we actually use gomock is interesting as we rely on `tools.go`, which runs the vendored `gomock` version, so we actually don't need to install it.
- I unexported all non required exportable things (like the runner) to not make adopters use things that they shouldn't touch. To make some bits testable, I exported them in an `export_test.go` file. The latter is a trick to export unexported bits that can't be used by a normal go app as the go build tool IGNORES symbols defined in `_test.go` files. I did something similar with `internal` that can't be imported by other packages in general.
- All the files containing logic are at the root level of `sdk/component`. This was done to have a better UX for contributing as well as shorter import paths. This is a common thing done for OSS projects.
- I opted for writing a simple but effective env var lookup functionality which is ready to be extended with other types of env var. This keeps the dependencies minimal.
- We always inject `component_name`, `component_type` and `sdk_version` as log fields to help us debugging issues.

---

Attaching some sample logs from running a sample scanner:

```
{"time":"2024-10-24T13:39:11.767624+01:00","level":"DEBUG","msg":"running component...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner"}
{"time":"2024-10-24T13:39:11.767639+01:00","level":"DEBUG","msg":"preparing to execute component...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner"}
{"time":"2024-10-24T13:39:11.767661+01:00","level":"DEBUG","msg":"preparing to execute scan step...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner"}
{"time":"2024-10-24T13:39:11.767665+01:00","level":"INFO","msg":"Scanning.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner"}
{"time":"2024-10-24T13:39:11.76769+01:00","level":"DEBUG","msg":"scan step completed!","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767692+01:00","level":"DEBUG","msg":"preparing to execute transform step...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767695+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767699+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.7677+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767702+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767725+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767728+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.76773+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767732+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767733+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767766+01:00","level":"INFO","msg":"Transforming.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10}
{"time":"2024-10-24T13:39:11.767771+01:00","level":"DEBUG","msg":"transform step completed!","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767774+01:00","level":"DEBUG","msg":"preparing to execute store step...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767776+01:00","level":"INFO","msg":"Storing.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767822+01:00","level":"DEBUG","msg":"store step completed!","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767825+01:00","level":"DEBUG","msg":"component has completed successfully!","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767827+01:00","level":"DEBUG","msg":"component done! Preparing to exit...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767835+01:00","level":"DEBUG","msg":"component done, exiting...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767869+01:00","level":"DEBUG","msg":"gracefully shutting down component...","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767875+01:00","level":"INFO","msg":"Closing scanner.","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
{"time":"2024-10-24T13:39:11.767877+01:00","level":"DEBUG","msg":"gracefully shutdown component successfully!","sdk_version":"0.0.1-alpha","component_name":"sample-scanner","component_type":"scanner","num_raw_findings":10,"num_parsed_findings":10,"num_findings_match":true}
```

---

Forgive me if the examples are a bit dummy but we can make a better job there when we have the storage later in place 🙏🏻


sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+* running the components steps in a predictable and reliable way
+* handle intricacies like cancellations and graceful shutdown
+* taking care of logging and panic handling
+* reporting common metrics to track what your component is doing
+
+You can customise a component using the following environment variables:
+
+| Environment Variable       | Type   | Required | Possible Values                |
+|----------------------------|--------|----------|--------------------------------|
+| SMITHY\_COMPONENT\_NAME       | string | yes      | -                              |
```
northdpole, line 26: <3 :heart:

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+	"log"
+	"time"
+
+	"github.com/smithy-security/smithy/sdk/component"
+)
+
+type sampleTarget struct{}
+
+func (s sampleTarget) Close(ctx context.Context) error {
+	// Close your component here!
```
northdpole, line 63:
```suggestion
	// Close your component here! 
	// Useful for cleanup actions.
	// For example if you are cloning a git repository, perhaps you want to delete paths that aren't in the sparseCheckout list
```

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+	return nil
+}
+
+func (s sampleTarget) Prepare(ctx context.Context) error {
+	// Prepare the target here!
```
northdpole, line 68:
```suggestion
	// Prepare the target here!
	// This is the main execution method of the Target component type. Here you need to implement your logic.
	// For example for a target component that clones a repository here is where you setup your arguments and call git clone
```

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+	return &ocsf.VulnerabilityFinding{}, nil
+}
+
+func (s sampleScanner) Close(ctx context.Context) error {
+	// Close your component here!
+	return nil
+}
+
+func (s sampleScanner) Store(ctx context.Context, findings []*ocsf.VulnerabilityFinding) error {
+	// Store your findings here!
```
northdpole, line 117: isn't the SDK taking care of this? does each component need to re-implement store?

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+	return nil
+}
+
+func (s sampleScanner) Scan(ctx context.Context) ([]component.Unmarshaler, error) {
+	// Scan a target and return a payload here!
```
northdpole, line 122: if I am running a binary, e.g. gosec, do i have to shell out, run the binary and return? can i call the binary in a previous step?

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+	return nil, nil
+}
+
+func (s sampleScanner) Transform(ctx context.Context, payload component.Unmarshaler) (*ocsf.VulnerabilityFinding, error) {
+	// Transform your payload to ocsf format!
```
northdpole, line 127: could you explain the difference between transform and unmarshall?

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+type sampleEnricher struct{}
+
+func (s sampleEnricher) Close(ctx context.Context) error {
+	// Close your component here!
+	return nil
+}
+
+func (s sampleEnricher) Read(ctx context.Context) ([]*ocsf.VulnerabilityFinding, error) {
+	// Read the prepared vulnerability reports here.
```
northdpole, line 167: Does this Read from the database or any intermediate data store?
If so, in reality does it call the sdk method "Read"?
If so, does it make sense to be called Load?

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+	return make([]*ocsf.VulnerabilityFinding, 0, 10), nil
+}
+
+func (s sampleEnricher) Update(ctx context.Context, findings []*ocsf.VulnerabilityFinding) error {
```
northdpole, line 171: could you provide an example of what this does? is it a "WriteResults" method?

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+	}
+}
+```
+
+#### Filter
+
+A `Filter` component allows to filter out some vulnerability findings based on
+arbitrary criteria.
+
+For example, you might want to filter out vulnerabilities on a specific path in a repository.
```
northdpole, line 196:
```suggestion
For example, you might want to filter out vulnerabilities on a specific path in a repository (useful for tests). Or you may want to ignore findings below a certain threshold.
```

northdpole (approved):

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+* reporting common metrics to track what your component is doing
+
+You can customise a component using the following environment variables:
+
+| Environment Variable       | Type   | Required | Possible Values                |
+|----------------------------|--------|----------|--------------------------------|
+| SMITHY\_COMPONENT\_NAME       | string | yes      | -                              |
+| SMITHY\_LOGGING\_LOG\_LEVEL    | string | false    | info, debug, warn, error       |
+
+Or you can use the supplied options `RunOption`, like in this example:
```
dlicheva, line 29: It might be clearer to say that "RunTarget" has a RunnerOptions param where the user can customise how a component runs. It would be also good to say what options are available.

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+)
+```
+
+### Components
+
+#### Target
+
+A `Target` component should be used to prepare a target for scanning.
+
+For example, cloning a repository and make it available for a `Scanner` to scan.
```
dlicheva, line 45: It would be clearer to give an example of an existing component. e.g. "Git-clone" is a target component because it clones a repository and prepares it for scanning.

sdk/README.md
dlicheva, line 63: What would be an example of closing a component?

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+)
+
+type (
+	sampleScanner struct{}
+
+	sampleRawVuln struct{}
+)
+
+func (s sampleRawVuln) Unmarshal() (*ocsf.VulnerabilityFinding, error) {
+	// Tell us how to convert your payload to ocsf format here!
```
dlicheva, line 107: That's a brilliant explanation :heart:

dlicheva: Should we add a comment with a link to the [OCSF](https://docs.aws.amazon.com/security-lake/latest/userguide/open-cybersecurity-schema-framework.html) website?

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+```go
+component.RunTarget(
+	ctx, 
+	sampleTarget{}, 
+	component.RunnerWithComponentName("sample-target"),
+)
+```
+
+### Components
```
dlicheva, line 39: The explanations are great, and I think it would be even clearer if we give an example of each component type. e.g. when we start describing Targets, say that git-clone is a target component, or when we start describing Reporters, say that the Slack reporter is one example. So people who learn from examples can have an easier time by reading the component code.

sdk/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+`component` makes it easy for you to leverage the default logger in your business logic.
+
+You can access the logger anytime using `component.LoggerFromContext(ctx)`.
+
+For example:
+
+```go
+func (s sampleEnricher) Update(ctx context.Context, findings []*ocsf.VulnerabilityFinding) error {
+	component.LoggerFromContext(ctx).Info("Preparing to update findings")
```
dlicheva, line 305: Nice :raised_hands:

sdk/component/enricher.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// RunEnricher runs an enricher after initialising the run context.
+func RunEnricher(ctx context.Context, enricher Enricher, opts ...RunnerOption) error {
+	return run(
+		ctx,
+		func(ctx context.Context) error {
+			logger := LoggerFromContext(ctx).With(logKeyComponentType, "enricher")
+
+			logger.Debug("preparing to execute enricher component...")
```
dlicheva, line 15: What do you think about adding the component name to the log? Might be useful if someone is debugging multiple components and trying to find out which one fails.

dlicheva (commented):
Nice work @andream16 ! I've just added some minor suggestions in the readme, but what I understand of the SDK looks great!

sdk/README.md
andream16: For example closing the git client :) I can add some explanation in the comments

sdk/README.md
andream16: Yeah, perhaps this can be unified - let me see it a bit

sdk/README.md
andream16: A `Scan` should act on a prepared `Target`, which was prepared by a previous `Target` component - as discussed

It should be ok whether the preparation is done via `Target` or script before this component imo

sdk/README.md
andream16: I don't feel strongly about the name

This will fetch ready VulnerabilityFindings from some source. By default this will be from SQLite and on SAAS from another data source - but one can override it - for example, reading from s3 and such

sdk/component/enricher.go
andream16: This is done by default for every runner [here](https://github.com/smithy-security/smithy/blob/5d6f911170ba22fbb16a6cea58f3e759ad61c6b5/sdk/component/runner.go#L73) :)

That's the power of centralising into a single runner!

sdk/component/runner.go
```diff
@@ -0,0 +1,202 @@
...
+		ContextWithLogger(ctx, logger),
+		syscall.SIGINT,
+		syscall.SIGTERM,
+		syscall.SIGQUIT,
+		syscall.SIGABRT,
+		syscall.SIGKILL,
+	)
+	defer cancel()
+
+	g, ctx := errgroup.WithContext(ctx)
```
ptzianos, line 92: this whole section looks like it's more complex than it's needed to be because of the use of an errgroup

I would expect that we would iterate over the stages of a workflow and for each stage a set of goroutines would be generated with each one sequentially executing the steps listed in a component, and all these goroutines would be added to a waitgroup. if any of the component steps throws an error the Close function is invoked to do whatever cleanup is needed.

using an errgroup means that we need to add all the channels for more synchronisation. instead we could just use a shared error with a mutex to protect a join function call, so that at the end we can have all the errors in one place.

components should be allowed to finish their job even if one of them fails, unless the user sends a kill signal to the process

sdk/component/target_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Prepare(gomock.Any()).
+			Return(errPrepare)
+		mockTarget.
+			EXPECT().
+			Close(gomock.Any()).
+			Return(nil)
+
+		err := runTargetHelper(t, ctx, mockTarget)
+		require.Error(t, err)
+		assert.ErrorIs(t, err, errPrepare)
```
ptzianos, line 89: why use an assert here instead of require? leaving this comment here, but it's present in various places

sdk/component/target.go
```diff
@@ -0,0 +1,31 @@
...
+		ctx,
+		func(ctx context.Context) error {
+			logger := LoggerFromContext(ctx).With(logKeyComponentType, "target")
+
+			logger.Debug("preparing to execute target component...")
+			logger.Debug("preparing to execute preparation step...")
+
+			if err := target.Prepare(ctx); err != nil {
+				logger.With(logKeyError, err.Error()).Error("preparing target failed")
+				return fmt.Errorf("could not prepare target: %w", err)
```
ptzianos, line 20: please use `github.com/go-errors/errors` instead of fmt.Errorf for better consistency

sdk/component/target.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// RunTarget runs a target after initialising the run context.
+func RunTarget(ctx context.Context, target Target, opts ...RunnerOption) error {
+	return run(
+		ctx,
+		func(ctx context.Context) error {
+			logger := LoggerFromContext(ctx).With(logKeyComponentType, "target")
+
+			logger.Debug("preparing to execute target component...")
```
ptzianos, line 15: nit: two logger errors over here are not adding more info, one should be enough

sdk/component/target.go
```diff
@@ -0,0 +1,31 @@
...
+			logger.Debug("preparing to execute preparation step...")
+
+			if err := target.Prepare(ctx); err != nil {
+				logger.With(logKeyError, err.Error()).Error("preparing target failed")
+				return fmt.Errorf("could not prepare target: %w", err)
+			}
+
+			logger.Debug("preparation step completed!")
```
ptzianos, line 23: nit: same comment applies for these two logging messages as above

sdk/README.md
andream16: Updated

sdk/README.md
andream16: Done by providing a default store

sdk/README.md
andream16: Updated

sdk/component/conf.go
```diff
@@ -0,0 +1,107 @@
...
+		Logging      RunnerConfigLogging
+		PanicHandler PanicHandler
+		// TODO: add MetricsHandler.
+		// TODO: add TracingHandler.
+		// TODO: add ProfilingHandler.
+	}
+
+	// RunnerConfigLogging contains the configuration related with the runner logger.
+	RunnerConfigLogging struct {
+		Level string
```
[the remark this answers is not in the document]
andream16: Done

sdk/component/target_test.go
andream16: Nothing bad in using an assert here as the next steps in the test case don't depend on it's success - also there are no steps next!

In general I find using `require` great for when we MUST assert that something is valid before moving on - for example, checking that a slice has 2 elements before doing assertions on those two elements

`assert` on the other end, doesn't stop the test if it fails and it's great to be used when we have multiple assertions in a block that wouldn't cause other issues in the test case if fail. The feedback loop of development using assert for failing test cases is faster than `require`.

If you try this example you'll get what I mean!

```go
func TestAssert(t *testing.T) {
	users := []user{
		{name: "Andrea", email: "andrea@smithy.security.com"},
		{name: "Pavlos", email: "pavlos@smithy.security.com"},
	}

	require.Len(t, users, 2)
	assert.Equal(t, "Diana", users[0])
	assert.Equal(t, "Spyros", users[1])
}

func TestRequire(t *testing.T) {
	users := []user{
		{name: "Andrea", email: "andrea@smithy.security.com"},
		{name: "Pavlos", email: "pavlos@smithy.security.com"},
	}

	require.Len(t, users, 2)
	require.Equal(t, "Diana", users[0])
	require.Equal(t, "Spyros", users[1])
}
```

sdk/component/target.go
andream16: Since go 1.13 `go-errors` is not necessary as `fmt.Errorf` and `errors` take care of things that `go-errors` covered https://go.dev/blog/go1.13-errors.

Actually Unwrap works better with `fmt.Errorf` for reference https://go.dev/play/p/U896OQVWlW4

andream16:
Closed in favour of https://github.com/smithy-security/smithy/pull/452

PR #449, opened 2024-10-25
Bugfix/scorecard producer

northdpole:
you can test this by running the example workflow for scorecard


components/producers/ossf-scorecard/main.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
```
andream16, line 6: I see that we're using both `log` and `slog`, can we just use `slog`?

components/producers/ossf-scorecard/main.go
```diff
@@ -36,19 +37,22 @@ func main() {
...
 
 func parseIssues(out *ScorecardOut) []*v1.Issue {
+	slog.Info("read ", slog.Int("numChecks", len(out.Checks)))
 	issues := []*v1.Issue{}
 	repo := out.Repo.Name
 	commit := out.Repo.Commit
 	for _, r := range out.Checks {
-		desc, _ := json.Marshal(r)
+		var desc string
+		for _, deet := range r.Details {
```
andream16, line 46: Should we skip adding a new line on the last entry of the slice?

components/producers/ossf-scorecard/main.go
```diff
@@ -72,36 +76,37 @@ func scorecardToSmithySeverity(score float64) v1.Severity {
...
+	Reason        string   `json:"reason,omitempty"`
+	Name          string   `json:"name,omitempty"`
+	Documentation Docs     `json:"documentation,omitempty"`
 }
 
 // Docs represents a ScoreCard "docs" section.
 type Docs struct {
-	URL   string
-	Short string
+	URL   string `json:"url,omitempty"`
```
andream16, line 98: This can probably be a [`url.URL`](https://pkg.go.dev/net/url#URL)

examples/pipelines/scorecard-project/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+nameSuffix: -scorecard-project
+components:
+  - pkg:helm/smithy-security-oss-components/base
+  - components/producers/ossf-scorecard
```
andream16, line 7: This needs to be updated to `pkg` as well :)

components/producers/ossf-scorecard/main.go
```diff
@@ -36,19 +37,22 @@ func main() {
 }
 
 func parseIssues(out *ScorecardOut) []*v1.Issue {
+	slog.Info("read ", slog.Int("numChecks", len(out.Checks)))
 	issues := []*v1.Issue{}
 	repo := out.Repo.Name
 	commit := out.Repo.Commit
 	for _, r := range out.Checks {
-		desc, _ := json.Marshal(r)
+		var desc string
```
andream16, line 45: Why did we change this from json?

northdpole: because it's going into a generic string, dumping json into a string field is usually ill-advised

components/producers/ossf-scorecard/main.go
northdpole: there's no slog.Fatal

components/producers/ossf-scorecard/main.go
northdpole: sometimes it's empty and a `url.URL`  crashes if it doesn't find a value

andream16 (approved):

PR #451, opened 2024-10-28
Add component release channels

northdpole:


ptzianos:
we should rework this idea on a new branch and with more thought into it

PR #455, opened 2024-10-30
Implement embedded storage backend reader and writer - Part 1

andream16:
Preparing to finish up the implementation with SQLite by:
- setting up things for SQLite in term of dependencies
- support reading the supplied `workflow_id` from the environment and leverage it for storing purposes in `Read`, `Update` and `Write`. We need such ID to connect the dots when writing and updating findings in the DB.

---

We agreed with @ptzianos to leverage the workflow_id from env var for now. We can cover the instance_id separately :)

---

I added a reusable uuid package which wraps google's UUID. Why? Because we're in charge of swapping the implementation and the interfaces don't have 3rd party dependencies.

---

**What did I do?**
- Added SQLite and sample TO BE IMPLEMENTED code
- Added reusable UUID package
- Passing workflow_id around where needed - also to be tracked in logs
- Refreshed examples, runners and tests to use the workflow_id

---

**In the next PR**
- Implement SQLite layer
- Write integration tests for SQLite layer


sdk/component/runner.go
```diff
@@ -69,6 +69,7 @@ func run(
 		logger = r.
 			config.Logging.Logger.
 			With(logKeySDKVersion, conf.SDKVersion).
+			With(logKeyWorkflowID, conf.WorkflowID.String()).
```
andream16, line 72: Now we can also track this bit.

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, fmt.Errorf("could not open sqlite db: %w", err)
+	}
+
+	return &manager{
+		db:     db,
+		dbName: dbName,
+	}, nil
+}
+
+// TODO - implement me.
```
andream16, line 59: To be done in the next PR.

sdk/component/internal/storer/local/local.go
```diff
@@ -14,22 +15,38 @@ func NewStoreManager() (*storeManager, error) {
 	return &storeManager{}, nil
 }
 
+// TODO - implement me.
```
andream16, line 18: This is like an adaptor layer - will implement it in the next PR.

sdk/component/env.go
```diff
@@ -7,13 +7,15 @@ import (
...
+
+	"github.com/smithy-security/smithy/sdk/component/internal/uuid"
 )
 
 type (
 	// parseableEnvTypes represents the types the parser is capable of handling.
 	// TODO: extend with slices if needed.
 	parseableEnvTypes interface {
-		string | bool | int | uint | int64 | uint64 | float64 | time.Duration | time.Time | url.URL
+		string | bool | int | uint | int64 | uint64 | float64 | time.Duration | time.Time | url.URL | uuid.UUID
```
andream16, line 18: Will support UUIDs out of the box from env var

northdpole (approved):

PR #457, opened 2024-10-31
workaround defectdojo expecting a non-empty scanstarttime for their e…

### LINKED ISSUE #456: DefectDojo crashes if the engagement time is 0
We can set the scanstarttime to 0 in case of tests or when running individual components so let's work around this bug by checking for the value and setting it to now() if it's 0

northdpole:
…ngagments
fix #456

### DIFF
--- a/components/consumers/defectdojo/main.go
+++ b/components/consumers/defectdojo/main.go
@@ -3,7 +3,11 @@ package main
 import (
 	"flag"
 	"log"
+	"log/slog"
 	"strconv"
+	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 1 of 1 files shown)


components/consumers/defectdojo/main.go
```diff
@@ -27,6 +30,14 @@ var (
 	issueTemplate          string
 )
 
+func getEngagementTime(engagementTime *timestamppb.Timestamp, scanID string) string {
+	if time.Time.IsZero(engagementTime.AsTime()) {
```
andream16, line 34: Should this be done on `engagementTime` directly?

`engagementTime.AsTime().IsZero()`

PR #459, opened 2024-10-31
Implement embedded storage backend reader and writer - Part 2

andream16:
Very simple local SQLite setup to put together a POC for Tuesday.

Right now I decided to set it up to convert the ocsf proto to JSON and storing it alongside a workflow id. This is done only for building the POC quickly - it shouldn't be the final solution.

We can work together when you are back @ptzianos to organise the data in the DB in a way that makes sense.

I didn't spend time on migrations as we might change the structure and we'll run them from the local potato engine for the POC demo.

---

**Next steps**:
- Merge this and start working on local potato engine for the POC.
- Discuss with the team about how to organise this data in a better way.


sdk/component/storer.go
```diff
@@ -1,20 +1,26 @@
...
 
 func isAllowedStoreType(st storeType) bool {
 	return st == storeTypeLocal
 }
 
-// newStore - TODO - implement in another PR.
-func newStorer(storeType storeType) (Storer, error) {
-	localMgr, _ := local.NewStoreManager()
-	return localMgr, nil
+func newStorer(conf runnerConfigStorer) (Storer, error) {
```
andream16, line 17: To be extended for SAAS in future with `remote` type support.

sdk/component/examples/enricher/main.go
```diff
@@ -22,18 +20,7 @@ func main() {
...
-	}
-
-	if err := component.RunEnricher(
-		ctx,
-		sampleEnricher{},
-		component.RunnerWithComponentName("sample-enricher"),
-		component.RunnerWithStorer("local", storageManager),
-		component.RunnerWithWorkflowID(uuid.New()),
-	); err != nil {
+	if err := component.RunEnricher(ctx, sampleEnricher{}); err != nil {
```
andream16, line 23: Simpler to leverage a .env file rather than passing options.

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -50,31 +77,174 @@ func NewManager(dbPath string, dbName string) (*manager, error) {
...
+	}
+
+	return mgr, nil
 }
 
-// TODO - implement me.
-func (m *manager) Read(ctx context.Context) ([]*ocsf.VulnerabilityFinding, error) {
-	return nil, nil
+// Validate. TODO - implement.
+func (m *manager) Validate(*ocsf.VulnerabilityFinding) error {
```
andream16, line 96: As mentioned - let's work on this separately.

sdk/component/internal/storer/local/sqlite/export_test.go
```diff
@@ -0,0 +1,27 @@
+package sqlite
+
+import (
+	"fmt"
+)
+
+// CreateTable is used to create a table in testing settings.
+func (m *manager) CreateTable() error {
```
andream16, line 8: To be removed when we have a better migrations story post POC and discussion with the team about data structure.

sdk/component/internal/storer/local/sqlite/sqlite_test.go
```diff
@@ -0,0 +1,218 @@
...
+		component.Closer
+		component.Reader
+		component.Updater
+		component.Writer
+
+		CreateTable() error
+	}
+
+	ManagerTestSuite struct {
+		suite.Suite
```
andream16, line 37: Using https://pkg.go.dev/github.com/stretchr/testify/suite for nice bdd like setup

sdk/component/internal/storer/storer.go
```diff
@@ -0,0 +1,5 @@
+package storer
+
+import "errors"
+
+var ErrNoFindingsFound = errors.New("no findings found")
```
andream16, line 5: This lives here so it's not tightly coupled to the local db but can be swapped with the remote one.

sdk/component/internal/storer/local/sqlite/export_test.go
```diff
@@ -0,0 +1,27 @@
...
+	stmt, err := m.db.Prepare(`
+		CREATE TABLE finding (
+			id INTEGER PRIMARY KEY AUTOINCREMENT UNIQUE,
+			workflow_id UUID NOT NULL UNIQUE ,
+			findings TEXT NOT NULL,
```
andream16, line 13: SQLite doesn't support JSONB

northdpole (approved):

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -3,42 +3,69 @@ package sqlite
...
 	ErrInvalidConstructor struct {
 		argName string
 		reason  string
 	}
 )
 
+// ManagerWithClock allows customising manager's clock.
+func ManagerWithClock(clock clockwork.Clock) managerOption {
+	return func(m *manager) error {
+		if clock == nil {
```
ptzianos, line 50: I think this check should be outside of the function so that it fails immediately instead of failing when the function is invoked.

sdk/component/internal/storer/local/sqlite/sqlite.go
ptzianos: can you make a ticket for this please to track it and add the ticket number in the comment?

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -3,42 +3,69 @@ package sqlite
...
-func NewManager(dbPath string, dbName string) (*manager, error) {
-	switch {
-	case dbPath == "":
-		return nil, ErrInvalidConstructor{
-			argName: "db path",
-			reason:  errInvalidConstructorEmptyReason,
-		}
-	case dbName == "":
+func NewManager(dbPath string, dbName string, opts ...managerOption) (*manager, error) {
+	if dbName == "" {
```
ptzianos, line 68: we should also check the `dbPath` over here that it's not empty and has a valid pattern

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -50,31 +77,174 @@ func NewManager(dbPath string, dbName string) (*manager, error) {
...
 
-// TODO - implement me.
-func (m *manager) Write(ctx context.Context, findings []*ocsf.VulnerabilityFinding) error {
+// Read finds Vulnerability Findings by workflowID.
+// It returns storer.ErrNoFindingsFound is not vulnerabilities were found.
+func (m *manager) Read(ctx context.Context, workflowID uuid.UUID) ([]*ocsf.VulnerabilityFinding, error) {
+	stmt, err := m.db.PrepareContext(ctx, `
+		SELECT (findings) 
+			FROM finding 
+			WHERE workflow_id = :workflow_id
```
ptzianos, line 106: looking at the code again, I think that what we actually mean `instance_id` instead of `workflow_id`. instance id is the UUID of each execution of a workflow. the workflow itself doesn't have an ID, it could have a name or some other identifier, but in the context of the OSS smithy it will be a file with the stages in the local filesystem

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -50,31 +77,174 @@ func NewManager(dbPath string, dbName string) (*manager, error) {
...
+			sql.Named(columnNameWorkflowID.String(), workflowID.String()),
+		).
+		Scan(&jsonFindingsStr); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, fmt.Errorf("%s: %w", workflowID.String(), storer.ErrNoFindingsFound)
+		}
+		return nil, fmt.Errorf("could not select findings: %w", err)
+	}
+
+	if err := stmt.Close(); err != nil {
```
ptzianos, line 126: this should also be moved into a defer function

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -50,31 +77,174 @@ func NewManager(dbPath string, dbName string) (*manager, error) {
...
+			FROM finding 
+			WHERE workflow_id = :workflow_id
+		;
+	`)
+	if err != nil {
+		return nil, fmt.Errorf("could not prepare select statement: %w", err)
+	}
+
+	var jsonFindingsStr string
+	if err := stmt.
```
ptzianos, line 114: I'm not a huge fan of this format because the err is checked 4 lines down and it's kinda lost. I'd rather have the err checked in a separate if statement

sdk/component/internal/storer/local/sqlite/sqlite_test.go
```diff
@@ -0,0 +1,218 @@
...
+	defer cancel()
+
+	mts.t.Run("given an empty database, when I look for findings, I get none back", func(t *testing.T) {
+		resFindings, err := mts.manager.Read(
+			ctx,
+			workflowID,
+		)
+		require.Error(mts.t, err)
+		assert.ErrorIs(t, err, storer.ErrNoFindingsFound)
+		assert.Len(mts.t, resFindings, 0)
```
ptzianos, line 154: why mix `assert` and `require` in these tests?

sdk/component/conf.go
```diff
@@ -19,10 +19,15 @@ const (
 	errReasonCannotBeNil      = "cannot be nil"
 
 	// Env vars.
-	envVarKeyComponentName   = "SMITHY_COMPONENT_NAME"
-	envVarKeyWorkflowID      = "SMITHY_WORKFLOW_ID"
+	// -- BASE
+	envVarKeyComponentName = "SMITHY_COMPONENT_NAME"
```
ptzianos, line 23: this should be defined by the component itself in the entrypoint not an environment variable

sdk/component/conf.go
```diff
@@ -19,10 +19,15 @@ const (
...
+	envVarKeyWorkflowID    = "SMITHY_WORKFLOW_ID"
+	// -- LOGGING
 	envVarKeyLoggingLogLevel = "SMITHY_LOG_LEVEL"
-	envVarKeyBackedStoreType = "SMITHY_BACKEND_STORE_TYPE"
+	// -- STORE
+	envVarKeyBackendStoreType        = "SMITHY_BACKEND_STORE_TYPE"
```
ptzianos, line 28: these environment variables could be simplified into one environment variable with a URI like `sqlite:///home/user/bla/smithy.db`

sdk/component/storer.go
```diff
@@ -1,20 +1,26 @@
...
+	"github.com/smithy-security/smithy/sdk/component/internal/storer/local/sqlite"
+)
 
 type storeType string
 
-const (
-	storeTypeTest  storeType = "test"
-	storeTypeLocal storeType = "local"
-)
+const storeTypeLocal storeType = "local"
```
ptzianos, line 11: I think this should be an enum so that we can ensure the validity of the options. we are using the https://github.com/abice/go-enum in other places to make sure that the enumeration is actually useful and safe

sdk/component/conf.go
```diff
@@ -228,42 +239,59 @@ func newRunnerConfig() (*RunnerConfig, error) {
...
+		SDKVersion:    sdk.Version,
+		WorkflowID:    workflowID,
+		Logging: RunnerConfigLogging{
+			Level:  RunnerConfigLoggingLevel(logLevel),
+			Logger: logger,
+		},
+		PanicHandler: panicHandler,
+	}
 
 	if st != "" {
```
ptzianos, line 258: we should use a better name for the storer implementation and we should also return immediately if it's an empty string. I think using an enum generator would help in this place because it would prevent us from ever having to deal with a random string value but we will always have to deal with enum values

sdk/component/internal/storer/local/sqlite/export_test.go
```diff
@@ -0,0 +1,27 @@
...
+			created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+			updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+		);
+	`)
+	if err != nil {
+		return fmt.Errorf("could not prepare statement for creating table: %w", err)
```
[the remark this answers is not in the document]
andream16: Since go 1.13 `errors` and `fmt.Errorf` are sufficient to cover what `go-errors` used to do https://go.dev/play/p/U896OQVWlW4

I'd prefer to not bring in an extra dependency if possible!

Let me know what you think

sdk/component/internal/storer/local/sqlite/sqlite.go
andream16: The one outside can't return an error as it returns an option - commonly done in the functional options pattern :)

sdk/component/internal/storer/local/sqlite/sqlite_test.go
andream16: Usually require should be used for something that must not fail to move on with further assertions. A require failure will halt the test case

Assertions can instead fail and the test case can keep going on

My rule of thumb is:
- use require to stop the world and don't have noisy tests - i.e. a panic caused by accessing a slice which is empty
- use assert when it's safe to keep the test case going - it's generally nice to use when you have multiple assertions so that you can immediately see all of them failing - with require this would lead to having to fix them one at a time to discover the issues - with assert it's only one run - this is very good on ci usually

sdk/component/conf.go
andream16: Why is that? Can you elaborate?

sdk/component/internal/storer/local/sqlite/sqlite.go
andream16: https://linear.app/smithy/issue/OCU-259/validate-component-input-and-output

sdk/component/conf.go
andream16: We're validating these contents here already https://github.com/smithy-security/smithy/blob/main/sdk/component/runner.go#L40!

sdk/component/conf.go
andream16: Took this offline - I can tackle it in a separate small pr if that's ok

andream16: https://linear.app/smithy/issue/OCU-265/leverage-component-name-as-argument-for-components

sdk/component/conf.go
```diff
@@ -228,42 +237,48 @@ func newRunnerConfig() (*RunnerConfig, error) {
...
+			"smithy.db",
+			withFallbackToDefaultOnError(true),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("could not lookup environment for '%s': %w", envVarKeyBackendLocalStoreDSN, err)
 		}
-		store, err = newStorer(storageType)
+
+		conf.storerConfig.localConf.dbDSN = localDBDSN
+		conf.storerConfig.enabled = true
```
ptzianos, line 274: what's this bool used for?

sdk/component/component.go
```diff
@@ -18,19 +18,19 @@ type (
 	// Reader allows reading vulnerability findings from a storage.
```
ptzianos, line 18: can you shorten the commit message and add more context in the commit message please?

sdk/component/conf.go
```diff
@@ -19,10 +19,14 @@ const (
...
-	envVarKeyComponentName   = "SMITHY_COMPONENT_NAME"
-	envVarKeyWorkflowID      = "SMITHY_WORKFLOW_ID"
+	// -- BASE
+	envVarKeyComponentName = "SMITHY_COMPONENT_NAME"
+	envVarKeyInstanceID    = "SMITHY_INSTANCE_ID"
+	// -- LOGGING
 	envVarKeyLoggingLogLevel = "SMITHY_LOG_LEVEL"
-	envVarKeyBackedStoreType = "SMITHY_BACKEND_STORE_TYPE"
+	// -- STORE
+	envVarKeyBackendStoreType     = "SMITHY_BACKEND_STORE_TYPE"
```
ptzianos, line 28: I think we should be able to infer the storage type from the store DSN. if it starts with 'sqlite' for example it will be SQLite while if starts with grpc it will be the remote implementation

sdk/component/conf.go
```diff
@@ -19,10 +19,14 @@ const (
...
+	envVarKeyBackendLocalStoreDSN = "SMITHY_BACKEND_LOCAL_STORE_DSN"
```
ptzianos, line 29: I don't see why the `local` substring is needed in these names. the DSN could be anything, either an implementation talking directly to a database or some other implementation using some gateway and the local substring makes them seem like they are only used for the local runner

sdk/component/storer.go
ptzianos: the string enums in go are a fotogun, for example you could do `storeType("bla")` and the compiler has no issue with this but the wrong value would be allowed to progress quite a bit throughout the code before (hopefully) a switch or an if will figure out that the value is incorrect. The problem is that when writing code that uses enums you make the (reasonable) assumption that there are only specific values that you deal with, not any value out there. But in Golang that's not the case, in contrast to pretty much every other programming language out there.

sdk/component/reporter_test.go
```diff
@@ -127,7 +127,7 @@ func TestRunReporter(t *testing.T) {
 				Return(nil),
 		)
 
-		err := runReporterHelper(t, ctx, workflowID, mockReporter, mockStore)
+		err := runReporterHelper(t, ctx, instanceID, mockReporter, mockStore)
 		require.Error(t, err)
```
ptzianos, line 131: all these instances should be `require.ErrorIs` instead of splitting it into 2 checks

sdk/component/internal/storer/local/sqlite/sqlite_test.go
```diff
@@ -0,0 +1,218 @@
...
+			require.Len(mts.t, copyFindings, 2)
+			copyFindings[0].Metadata.Version = newVersion
+			copyFindings[1].Metadata.Version = newVersion
+
+			err := mts.manager.Update(ctx, instanceID, copyFindings)
+			require.NoError(mts.t, err)
+
+			resFindings, err := mts.manager.Read(ctx, instanceID)
+			require.NoError(mts.t, err)
+			require.Len(mts.t, resFindings, 2)
```
ptzianos, line 207: this could be simplified to `require.EqualValues`

sdk/component/internal/storer/local/sqlite/sqlite_test.go
```diff
@@ -0,0 +1,218 @@
...
+	mts.t.Run("given two findings are present in the database, I should be able to retrieve them", func(t *testing.T) {
+		resFindings, err := mts.manager.Read(ctx, instanceID)
+		require.NoError(mts.t, err)
+		require.Len(mts.t, resFindings, 2)
+		assert.Equal(mts.t, findings, resFindings)
+	})
+
+	mts.t.Run("given a non existing instance id in the database, updating should fail", func(t *testing.T) {
+		err := mts.manager.Update(ctx, uuid.New(), findings)
+		require.Error(mts.t, err)
```
ptzianos, line 188: this could be simplified to `require.ErrorIs`

sdk/component/internal/storer/local/sqlite/sqlite_test.go
ptzianos: Yes but in this case this is adding verbosity without extra visibility. As in other places, this could be simplified to `require.ErrorIs(t, err, storer.ErrNoFindingsFound)` and then `require.Len(t, resFindings, 0)`. the last assertion is at the end of the test Run so it doesn't make a difference, but having a require will make the code more consistent

sdk/component/internal/storer/local/sqlite/export_test.go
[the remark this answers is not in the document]
ptzianos: The `github.com/go-errors/errors` is interesting because of 2 reasons:
1. it ensures that the code base is more consistent everywhere because you are not using 2 packages when interacting with errors (the std errors and fmt) but only one which is named errors
2. most important it adds stacktraces to errors which can then be printed from the entrypoint at the top level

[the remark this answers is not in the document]
andream16: I can look into this separately after merging these changes if you don't mind https://linear.app/smithy/issue/OCU-268/replace-errors-and-fmterrorf-utilisations-for-go-errors-in-sdk

I don't want to have more changes in the PR :)

PR #471, opened 2024-11-08
ocu-223 local example of a go-sec pipeline

andream16:
Adding a local example for a go-sec pipeline.

You can simply use it by injecting the provider .env file to run the main! That's it!


go.mod
```diff
@@ -10,7 +10,8 @@ require (
 	github.com/avast/retry-go/v4 v4.3.3
 	github.com/aws/aws-sdk-go v1.17.7
 	github.com/elastic/go-elasticsearch/v8 v8.3.0
-	github.com/go-errors/errors v1.4.2
+	github.com/go-errors/errors v1.5.1
```
andream16, line 13: Will remove these deps after packaging components.

examples/pipelines/golang-project/scanner.go
```diff
@@ -0,0 +1,256 @@
...
+					},
+				},
+			},
+		})
+	}
+
+	return vulns, nil
+}
+
+func (g *goSecScanner) runGoSec(ctx context.Context) error {
```
andream16, line 208: Using dockertest here just for the POC - we'll use gosec as a container in a real world scenario

northdpole (approved):

PR #473, opened 2024-11-13
Parsing correctly pURLs and prefixing targets.

andream16:
- Prefixing physical (files) targets with `file://`
- Correctly lookup for pURLs which were previously ignored
- Snyk-python and snyk-node pass more context about the language so that we can improve not well formatted pURLs


pkg/sarif/sarif.go
```diff
@@ -230,15 +237,15 @@ func addIssue(rules map[string]*sarif.ReportingDescriptor, issues []*v1.Issue, t
...
 		Confidence:  v1.Confidence_CONFIDENCE_UNSPECIFIED,
 	})
 
 	return issues
 }
 
-// levelToseverity transforms error, warning and note levels to high, medium and low respectively.
-func levelToseverity(level string) v1.Severity {
+// levelToSeverity transforms error, warning and note levels to high, medium and low respectively.
+func levelToSeverity(level string) v1.Severity {
```
andream16, line 248: Small typo.

northdpole (approved):

northdpole (approved):

pkg/sarif/sarif.go
```diff
@@ -261,6 +283,180 @@ func severityToLevel(severity v1.Severity) string {
...
+	return target, nil
+}
+
+func isLogicalLocation(loc *sarif.Location) bool {
+	return len(loc.LogicalLocations) > 0
+}
+
+// parseLogicalTargets parses all the targets found in the logical locations.
+// A logical target must be a valid pURL.
+func parseLogicalTargets(loc *sarif.Location, lang ...ExtraContextLanguage) ([]string, error) {
```
ptzianos, line 391: I don't understand this choice, why have the language be a slice of arbitrary length? this could easily be invoked with `parseLogicalTargets(someLocation, "javascript", "python")` and it would not throw an error. since the code can only handle one language every time we should allow only one language to be passed.

this piece of code would be simplified if it was a method of a struct that could hold the context of the code.
for example we could have `sarif.Parser` and define a method called `WithContext(someLanguage string)` to add context when we have it so that we can reduce the number of parameters needed when the publicly available methods are invoked instead of having one parameter percolate through three functions (`ToSmithy` -> `parseOut` -> `parseTargets`)

andream16: Right now it uses the first item in the slice, if defined

I did it to not break the existing contract for the `ToSmithy` parser

I thought of refactoring the code more heavily to do what you suggested but I'm not sure if it's worth the investment since we'll soon rework on this in the OSS

If you feel strongly about it I can make more changes

ptzianos: I looked really quick into how many components are using the sarif parser and it (https://github.com/search?q=repo%3Asmithy-security%2Fsmithy%20sarif.ToSmithy&type=code) looks there is 5-6 out there which makes me question how accurate they are at parsing results, but that's a discussion for another day. for now I don't think it will be a huge amount of changes if we add a language parameter to the `ToSmithy` function and just pass an empty string to the components that we are not interested atm. this will also not add too many changes to the current PR
I guess this piece of code will be completely refactored in order to integrate with the new SKD anyway

andream16: Sounds good - I passed the language and used `Unspecified` when not needed

The code handles both empty val and unspecified to understand that it should not be used.

andream16:
Closed in favour of https://github.com/smithy-security/smithy/pull/476.

PR #477, opened 2024-11-18
allow snyk node to be used for monorepos

ptzianos:
our current snyk node component can only be used for repositories where the root folder is the one containing the NodeJS code, however it can't be used if the folder structure is any different or if there are multiple subfolder containing NodeJS projects.

this PR fixes this by allowing the user to define multiple paths to be scanned, and uses a better algorithm for deciding which package manager to use to resolve their dependencies. it stores the findings in a way that the parser can use to make sure that it has the full context about where they were discovered

### DIFF
--- a/components/producers/snyk-node/task.yaml
+++ b/components/producers/snyk-node/task.yaml
@@ -11,50 +11,45 @@ spec:
   params:
   - name: producer-snyk-node-api-key
     type: string
+  - name: producer-snyk-node-directories
+    type: array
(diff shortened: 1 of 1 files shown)


andream16 (approved):

andream16 (approved):

northdpole:
this needs modification for working with a `package.lock` instead of yarn

dlicheva (approved):

PR #478, opened 2024-11-19
PDF consumer updates

dlicheva:
[Ticket](https://linear.app/smithy/issue/SAAS-730/component-pdf-improvements)
This PR fixes the following issues with the PDF consumer:

- only upload the PDF report, not the html
- avoid creating an html report in the component folder when running the test
- ensure each tool report is on a new page
- fix overflowing descriptions of vulnerabilities

### DIFF
--- a/components/consumers/pdf/main.go
+++ b/components/consumers/pdf/main.go
@@ -89,7 +89,7 @@ func main() {
 	}
 }
 
-func run(responses any, s3FilenamePostfix string, pw playwright.Wrapper, s3Wrapper s3client.Wrapper) error {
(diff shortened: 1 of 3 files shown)


components/consumers/pdf/main.go
```diff
@@ -140,6 +147,11 @@ func buildPdf(data any, pw playwright.Wrapper) (string, []byte, error) {
 		return "", nil, fmt.Errorf("could not generate pdf from page %s, err: %w", reportPage, err)
 
 	}
+
+	// delete the intermediate HTML file
+	if err := os.Remove(reportHTMLPath); err != nil {
```
andream16, line 152: ⛏️ I think we should log this error and keep going as it's not critical - more or less like the other check that you added to close the file 🤓

andream16 (approved):

PR #480, opened 2024-11-21
Implementing git-clone target

andream16:
Basic git-clone target built on top of the SDK.

Will look into packaging it after we have a few components from a sample pipeline in place.

---

Will eventually deprecate `new-components` - right now this is the fastest and cleanest thing we can do imo.


.github/workflows/test.yml
```diff
@@ -16,6 +16,11 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    services:
```
andream16, line 19: Like we did in SAAS to run integration tests on CI.

new-components/targets/git-clone/go.mod
```diff
@@ -0,0 +1,88 @@
...
+
+go 1.23.2
+
+require (
+	github.com/go-errors/errors v1.5.1
+	github.com/go-git/go-billy/v5 v5.5.0
+	github.com/go-git/go-git/v5 v5.12.0
+	github.com/jonboulle/clockwork v0.4.0
+	github.com/ory/dockertest/v3 v3.11.0
+	github.com/smithy-security/pkg/env v0.0.0-20241121114636-0915d48f6d71
```
andream16, line 11: Need to merge and tag  https://github.com/smithy-security/pkg/pull/1and update this dep.

new-components/targets/git-clone/cmd/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"time"
+
+	"github.com/go-errors/errors"
+	"github.com/smithy-security/smithy/sdk/component"
+
+	"github.com/smithy-security/smithy/new-components/targets/git-clone/pkg/git"
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
andream16, line 15: Sample timeout here if we clone forever. Let's customise this in future if needed.

new-components/targets/git-clone/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+ENV CGO_ENABLED=0
+ENV GOOS="linux"
+ENV GOARCH="amd64"
+RUN go build -ldflags "-s -w" -trimpath -o app ./cmd/main.go
+
+# Create a workspace to clone repos to.
+RUN mkdir -p /workspace
+
+# Used to actually run the binary in minimal image.
+FROM gcr.io/distroless/base-debian12:nonroot
```
andream16, line 29: I keep getting errors for missing .so(s) using scratch - if you have suggestions you are welcome!

new-components/targets/git-clone/docker-compose.yaml
```diff
@@ -0,0 +1,30 @@
...
+    platform: linux/amd64
+    env_file:
+      - .env.git-clone
+    volumes:
+      - ./testing/testdata/testrepo:/workspace
+    depends_on:
+      gitea:
+        condition: service_healthy
+  gitea:
+    image: gitea/gitea:latest
```
andream16, line 15: Setting up a local GitHub like backend to be used for local setup and testing https://about.gitea.com/.

new-components/targets/git-clone/testing/integration/git_test.go
```diff
@@ -0,0 +1,191 @@
...
+	gitCloneTargetResource *dockertest.Resource
+}
+
+func TestTestGitCloneSuite(t *testing.T) {
+	suite.Run(t, new(TestGitCloneSuite))
+}
+
+func (s *TestGitCloneSuite) SetupSuite() {}
+
+func (s *TestGitCloneSuite) SetupTest() {
```
andream16, line 37: Basically spinning up gitea and then the container cloning a sample repo.

northdpole (approved):

new-components/targets/git-clone/pkg/git/git.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+const errInvalidConfigurationReasonEmpty = "cannot be empty"
+
+type (
+	// Conf wraps the component configuration.
+	Conf struct {
+		ClonePath  string
+		RepoURL    string
+		BranchName string
```
northdpole, line 22: this is missing a ref, i understand this is an MVP let's just add a ticket to implement this whenever someone asks for it

new-components/targets/git-clone/pkg/git/git.go
```diff
@@ -0,0 +1,207 @@
...
+
+		ConfAuth ConfAuth
+	}
+
+	// ConfAuth contains authentication configuration.
+	ConfAuth struct {
+		Username    string
+		AuthEnabled bool
```
northdpole, line 30: the default git auth is private key, not necessary now but let's add a ticket for this please

new-components/targets/git-clone/pkg/git/git.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
+	"github.com/smithy-security/pkg/env"
+)
+
+const errInvalidConfigurationReasonEmpty = "cannot be empty"
+
+type (
+	// Conf wraps the component configuration.
+	Conf struct {
```
northdpole, line 19: this is also missing a set of dirs for git sparse checkout, useful for monorepos

andream16: Going to skip this for now as discussed in person

new-components/targets/git-clone/go.mod
andream16: Done

new-components/targets/git-clone/pkg/git/git.go
andream16: Changed to ref.

new-components/targets/git-clone/pkg/git/git.go
andream16: Done! https://linear.app/smithy/issue/OCU-272/support-private-keys-for-new-git-clone-component

new-components/targets/git-clone/pkg/git/git.go
```diff
@@ -0,0 +1,207 @@
...
+		AccessToken string
+	}
+
+	// Repository wraps a git repository to not leak deps.
+	Repository struct {
+		repo *git.Repository
+	}
+
+	// ErrInvalidConfiguration is used in the constructor.
+	ErrInvalidConfiguration struct {
```
ptzianos, line 40: why create a custom error and not just use an Errorf? I don't see the extra information being used anywhere

new-components/targets/git-clone/cmd/main.go
```diff
@@ -0,0 +1,40 @@
...
+	"time"
+
+	"github.com/go-errors/errors"
+
+	"github.com/smithy-security/smithy/sdk/component"
+
+	"github.com/smithy-security/smithy/new-components/targets/git-clone/pkg/git"
+)
+
+func main() {
```
ptzianos, line 15: why the two mains?

new-components/targets/git-clone/docker-compose.yaml
```diff
@@ -0,0 +1,30 @@
+services:
```
ptzianos, line 1: nit: can we move this file into a `testdata` folder so that it's clear what it's for?

new-components/targets/git-clone/.env.git-clone
```diff
@@ -0,0 +1,6 @@
+SMITHY_COMPONENT_NAME=git-clone
```
ptzianos, line 1: I'd rather this value is hardcoded in the a file with component variables, such as name, type etc.

new-components/targets/git-clone/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# These CGO_CPPFLAGS help preventing overflows.
+# Add a small overhead at compile time.
+ENV CGO_CPPFLAGS="-D_FORTIFY_SOURCE=2 -fstack-protector-all"
+# Makes memory exploitation harder.
+# Add a small overhead at compile time.
+ENV GOFLAGS="-buildmode=pie"
+ENV CGO_ENABLED=0
+ENV GOOS="linux"
+ENV GOARCH="amd64"
+RUN go build -ldflags "-s -w" -trimpath -o app ./cmd/main.go
```
ptzianos, line 23: can we move this into a makefile, so that it can be used both inside and outside of the container?

new-components/targets/git-clone/testing/integration/git_test.go
```diff
@@ -0,0 +1,191 @@
+package integration
```
ptzianos, line 1: I'd rather we use a file structure that's more standard for Golang with all the tests in the same package as the component, only split into a separate file with the `// +build !integration` and then invoked with `go test --tags=integration` and all the test related files in the `testdata` folder

ptzianos:
@andream16 this is really good, only some nits here and there. can we also add a `go build -ldflags "-X my/package/config.Version=1.0.0"` type of flag so that we can hardcode the component version in the binary?

new-components/targets/git-clone/pkg/git/target.go
```diff
@@ -0,0 +1,135 @@
...
+	if _, err := g.cloner.Clone(ctx); err != nil {
+		return errors.Errorf("could not clone repository: %w", err)
+	}
+
+	var (
+		endTime       = g.clock.Now()
+		cloneDuration = endTime.Sub(startTime)
+	)
+
+	logger.
```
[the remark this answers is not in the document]
andream16: TBF if it's in one line it's trickier to read for me, so just a preference :)

new-components/targets/git-clone/pkg/git/git.go
andream16: Not a big deal, I was thinking of leveraging the error using errors.As but didn't end up doing it - will simplify

new-components/targets/git-clone/pkg/git/git.go
```diff
@@ -0,0 +1,207 @@
...
+	if envLoader != nil {
+		envOpts = append(envOpts, env.WithLoader(envLoader))
+	}
+
+	clonePath, err := env.GetOrDefault(
+		"GIT_CLONE_PATH",
+		".",
+		append(envOpts, env.WithDefaultOnError(true))...,
+	)
+	if err != nil {
```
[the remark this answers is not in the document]
andream16: Fair, I can remove

new-components/targets/git-clone/docker-compose.yaml
andream16: I prefer to have it at root level so it's easy to discover and acquainted people with compose know immediately what's the deal :)

We don't copy it in the containers anyways

new-components/targets/git-clone/testing/integration/git_test.go
andream16: Fair, I can make this change but let's tackle running the integration tests separately if you don't mind 

Let's leave the same tag for now then we can invest a bit into splitting the execution

Now it runs as a normal test as DIND is on on CI anyway :)

new-components/targets/git-clone/.env.git-clone
andream16: Injected in the main as an option

andream16:
> @andream16 this is really good, only some nits here and there. can we also add a `go build -ldflags "-X my/package/config.Version=1.0.0"` type of flag so that we can hardcode the component version in the binary?

@ptzianos perhaps let's do this with packaging and inject this as an env var in the SDK lib?

We already have a customisable sdk version variable but we can also add a way to customise the component version from there

new-components/targets/git-clone/docker-compose.yaml
ptzianos: but it's only used for tests right? there's going to be a component.json file soon in there too, so I think ppl will get the wrong impression that we are using docker-compose under the hood

new-components/targets/git-clone/Dockerfile
ptzianos: yes. but I can see someone wanting to run it locally and in that case they would need to figure out how to build it themselves. having a Makefile with a build target means that we can invoke the same make target inside or outside the Dockerfile and get a binary

.github/workflows/test.yml
ptzianos: Are you sure we need this? In the SaaS repo for example I'm building container images without dind which means that the docker daemon is present

new-components/targets/git-clone/docker-compose.yaml
andream16: For testing we use `dockertest`, we're not using the `docker-compose.yaml` actually - we could use the `docker-compose.yaml` but then we'd lose some control on how the test is executed - most likely we'd have to also run the integration test in the `docker-compose` and that's always not great because of startup order etc

I added it so we can run things locally - from my exp it's useful to validate behaviour locally and manually :)

I think this is a good balance

andream16:
> > > @andream16 this is really good, only some nits here and there. can we also add a `go build -ldflags "-X my/package/config.Version=1.0.0"` type of flag so that we can hardcode the component version in the binary?
> > 
> > 
> > @ptzianos perhaps let's do this with packaging and inject this as an env var in the SDK lib?
> > We already have a customisable sdk version variable but we can also add a way to customise the component version from there
> 
> can you please create a ticket for this to revisit it some time this week when we start working on the component stuff

@ptzianos yes, done here https://linear.app/smithy/issue/OCU-273/inject-component-version-as-environment-variable-in-the-sdk

new-components/targets/git-clone/Dockerfile
andream16: Done!

ptzianos (approved):

PR #485, opened 2024-11-26
OCU-274 - Automigrate on sqlite

andream16:
- Replacing env functionalities with shared package.
- Automigrating on SQLite storage.
- Go mod tidy && go mod vendor


ptzianos (approved):

sdk/component/internal/storer/local/sqlite/sqlite.go
```diff
@@ -211,6 +215,28 @@ func (m *manager) Close(ctx context.Context) error {
 	return nil
 }
 
+// TODO: potentially leverage migrations here but this is simple enough for now for local setup.
```
dlicheva, line 218: Could you please write up a ticket on Linear with the next steps and link it here? If not relevant, this can just be a comment, not a todo.

sdk/component/conf.go
```diff
@@ -248,10 +249,10 @@ func newRunnerConfig() (*RunnerConfig, error) {
 
 		conf.storerConfig.storeType = storageType
 
-		dbDSN, err := fromEnvOrDefault(
+		dbDSN, err := env.GetOrDefault(
 			envVarKeyBackendStoreDSN,
 			"smithy.db",
```
dlicheva, line 254: Should we put all defaults into constants at the top of the file or somewhere cenralized?

sdk/component/internal/storer/local/sqlite/sqlite.go
andream16: Done :)

sdk/component/conf.go
andream16: Right now this is only in use here and we don't have other default vals for env vars so I think it would be more consistent as is

I agree that we can do what you suggest if we add more of them :)

PR #491, opened 2024-12-04
Implement findings storage gateway reader and writer - Part 1

andream16:
- Refactor local store with SQLite to be on by default and overwriteable with an existing option. This also simplifies the code quite a lot.
- Move UUID utils to root so it can be used without hoops.
- Move Local storage to root to simplify it.


andream16:
Closing in favour of https://github.com/smithy-security/smithy/pull/492

PR #492, opened 2024-12-05
Implement persistent findings storage - Part 1

andream16:
**How to review these changes**
- Review this PR
- Review https://github.com/smithy-security/smithy/pull/493
- Review https://github.com/smithy-security/smithy/pull/494
- Review https://github.com/smithy-security/smithy/pull/496
- Review https://github.com/smithy-security/saas/pull/806

In the following graph, you can get an idea of the incremental changes:

image

---

First step into having storage work over components running remotely or on containers.

This is mainly a preparation for the changes to come.

I'm setting up a To Be Implemented postgres remote storage. This will be finalised in the next PRs.

- Refactor local store with SQLite to be on by default and overwriteable with an existing option. This also simplifies the code quite a lot.
- Move UUID utils to accessible package so it can be used without hoops and can import it also for testing components.
- Move Local storage to accessible package to simplify it.
- Code changes for complying with the package locations
- Empty postgresql remote store ready to be implemented in part 2


sdk/README.md
```diff
@@ -25,12 +25,8 @@ You can customise a component using the following environment variables:
 | Environment Variable                | Type   | Required | Default                  | Possible Values          |
 |-------------------------------------|--------|-----------|--------------------------|--------------------------|
 | SMITHY\_COMPONENT\_NAME             | string | yes      | -                        | -                        |
-| SMITHY\_BACKEND\_STORE\_TYPE        | string | yes     | -                        | local, test, \*remote    |
```
andream16, line 28: This is simplified in the upcoming PR.

sdk/component/examples/enricher/.env
```diff
@@ -1,4 +1,3 @@
 SMITHY_COMPONENT_NAME=sample-enricher
 SMITHY_INSTANCE_ID=8d719c1c-c569-4078-87b3-4951bd4012ee
 SMITHY_LOG_LEVEL=debug
-SMITHY_BACKEND_STORE_TYPE=local
```
andream16, line 4: No longer necessary.

sdk/component/conf.go
```diff
@@ -110,9 +100,9 @@ func (rc *RunnerConfig) isValid() error {
 			FieldName: "panic_handler",
 			Reason:    errReasonCannotBeNil,
 		}
-	case rc.storerConfig.store == nil:
+	case rc.Storer == nil:
```
ptzianos, line 103: this check is not enough because an interface could be pointing to a Nil value but itself is not Nil. the correct check is something like this:
```golang
return rc.Storer == nil || (reflect.ValueOf(rc.Storer).Kind() == reflect.Ptr && reflect.ValueOf(rc.Storer).IsNil())
```
we can move this into an independent function so that it's used in other places too

sdk/component/conf.go
```diff
@@ -163,22 +153,15 @@ func RunnerWithInstanceID(id uuid.UUID) RunnerOption {
...
+func RunnerWithStorer(store Storer) RunnerOption {
 	return func(r *runner) error {
-		switch {
-		case !isAllowedStoreType(storeType(stType)):
-			return ErrRunnerOption{
-				OptionName: "store_type",
-				Reason:     errReasonUnsupportedValue,
-			}
-		case store == nil:
+		if store == nil {
```
ptzianos, line 158: same comment about Nil check for the storer

sdk/component/runner.go
```diff
@@ -36,6 +38,14 @@ func newRunner(opts ...RunnerOption) (*runner, error) {
 		}
 	}
 
+	if r.config.Storer == nil {
```
ptzianos, line 41: same comment about Nil check for storer

sdk/component/store/local/store.go
```diff
@@ -105,12 +113,12 @@ func (m *manager) Read(ctx context.Context, instanceID uuid.UUID) ([]*ocsf.Vulne
 	err = stmt.
 		QueryRowContext(
 			ctx,
-			sql.Named(ColumnNameInstanceId.String(), instanceID.String()),
+			sql.Named(LocalStoreColumnNameInstanceId.String(), instanceID.String()),
```
ptzianos, line 116: nit: this naming is a bit smurfy because it will result in something like:
```golang
localstore import "github.com/smithy-security/smithy/sdk/component/store/local"

localstore.LocalStoreColumnNameInstanceId
```

dlicheva (approved):

sdk/component/conf.go
andream16: Done

sdk/component/conf.go
andream16: Done

sdk/component/runner.go
andream16: Done

sdk/component/store/local/store.go
andream16: This is going away in the next PR regardless so I'd ignore it :)

PR #499, opened 2024-12-17
Open Sourcing V1 Types

andream16:
Moving useful V1 types from SAAS to OSS so we can start working on smithyctl.

All types have been moved except Instances, which are a SAAS only concern.

GORM and Kubernetes specific fields or helpers have been removed. These will be mapped in SAAS with an adapter layer.

In general, avoiding using pointers to make types and fields safer to use.

---

**Next steps:**
- Work on Smithyctl to create local workflows + packaging
- Make SAAS understand new OSS V1 types


dlicheva (approved):

smithyctl/types/v1/component.go
```diff
@@ -0,0 +1,44 @@
...
+	// ENUM(unknown, target, scanner, enricher, filter, reporter)
+	ComponentType string
+
+	// Component is a binary or a script that can be used in the context of a
+	// workflow to generate some useful result.
+	Component struct {
+		// Description describes what the component will do.
+		Description string
+		// ID is a unique identifier for a component.
+		ID uuid.UUID
```
ptzianos, line 18: I don't think we will need t a UUID for the component because it's not stored in a database and there is no central authority to check the uniqueness of this ID

smithyctl/types/v1/workflow.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/google/uuid"
+)
+
+type (
+	// Workflow represents a combination of
+	Workflow struct {
+		// Description described what the workflow is configured for.
+		Description string
+		// ID is the unique identifier of this workflow.
+		ID uuid.UUID
```
ptzianos, line 13: same with this ID

dlicheva (approved):

PR #674, opened 2025-01-07
OCU-281 - New bandit component

northdpole:
Works towards OCU-281


new-components/pkg/target/target.go
```diff
@@ -0,0 +1,9 @@
+package target
```
northdpole, line 1: an enum listing available target, could be used by validation to ensure we don't use random targets

andream16: I think that this is already available https://github.com/smithy-security/smithy/blob/main/proto/ocsf/ocsf_ext/finding_info/v1/finding_info.proto and can be imported from the local generated code in the SDK

new-components/pkg/test/util/fake_clock.go
```diff
@@ -0,0 +1,11 @@
+package testutil
+
+import (
+	"time"
+
+	"github.com/jonboulle/clockwork"
+)
+
+func FakeClock() clockwork.FakeClock {
+	return clockwork.NewFakeClockAt(time.Date(2024, 11, 1, 0, 0, 0, 0, time.UTC))
```
andream16, line 10: I'd avoid doing this as a static date can be harmful, I think one can simply take care of this in a test setup :)

new-components/pkg/test/util/test_transformer_wrapper.go
```diff
@@ -0,0 +1,202 @@
+package testutil
```
andream16, line 1: Sorry to be pedantic here, can we keep this in the needed component for now?

I'd like us to keep implementing the same bit a few times before it needs to be abstracted in another module

new-components/scanners/bandit/docker-compose.yaml
```diff
@@ -0,0 +1,12 @@
+services:
```
andream16, line 1: In the .env I see that we rely on postgres but I don't see a postgres container here 🤔

new-components/scanners/bandit/go.mod
```diff
@@ -0,0 +1,63 @@
+module github.com/smithy-security/smithy/new-components/scanners/bandit
```
andream16, line 1: What's the file called `new-components/scanners/bandit/go`?

new-components/scanners/bandit/internal/transformer/transformer.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, errors.Errorf("raw output file '%s' not found", b.rawOutFilePath)
+		}
+		return nil, errors.Errorf("failed to read raw output file '%s': %w", b.rawOutFilePath, err)
+	}
+
+	var results BanditOut
+	if err := json.Unmarshal(inFile, &results); err != nil {
+		log.Fatal(err)
```
andream16, line 156: Can we avoid using these statements and return an error so that the SDK can handle it correctly?

new-components/scanners/bandit/internal/transformer/transformer.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, errors.Errorf("failed to map code: %w", err)
+	}
+
+	return &ocsf.VulnerabilityFinding{
+		ActivityName: ptr.Ptr(ocsf.VulnerabilityFinding_ACTIVITY_ID_CREATE.String()),
+		ActivityId:   ocsf.VulnerabilityFinding_ACTIVITY_ID_CREATE,
+		CategoryUid:  ocsf.VulnerabilityFinding_CATEGORY_UID_FINDINGS,
+		ClassUid:     ocsf.VulnerabilityFinding_CLASS_UID_VULNERABILITY_FINDING,
+		ClassName:    ptr.Ptr(ocsf.VulnerabilityFinding_CLASS_UID_VULNERABILITY_FINDING.String()),
+
```
andream16, line 208: ⛏️ unnecessary newline here

new-components/scanners/bandit/internal/transformer/transformer.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, errors.New("invalid empty raw output file")
+	case t.targetType == "":
+		return nil, errors.New("invalid empty target type")
+	}
+
+	return &t, nil
+}
+
+// Transform transforms raw sarif findings into ocsf vulnerability findings.
+func (b *banditTransformer) Transform(ctx context.Context) ([]*ocsf.VulnerabilityFinding, error) {
```
andream16, line 141: 🔥

new-components/scanners/bandit/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+The component uses environment variables for configuration.
+
+It requires the component
+environment variables defined [here](https://github.com/smithy-security/smithy/blob/main/sdk/README.md#component) as well
+as the following:
+
+| Environment Variable     | Type   | Required | Default    | Description                                             |
+|--------------------------|--------|----------|------------|---------------------------------------------------------|
+| BANDIT\_RAW\_OUT\_FILE\_PATH  | string | yes      | -          | The path where to find the gosec report                 |
```
andream16, line 16: These need to be added to the .env :)

new-components/pkg/test/util/test_transformer_wrapper.go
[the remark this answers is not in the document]
andream16: Took this offline and we agreed on keeping things boring for now and copy pasting some code to manipulate or test ocsf input/outputs rather than creating shared utilities.

> Why is that?
If we create the utility now and start using it in a few components, we'll be aware only of a few use cases where such utilities are used. By doing that, we put ourselves in a place where we need to do more work when we discover that our utility doesn't cover another new use case and have to update both the utility and the components, with potential breaking changes.

For now, we should keep things boring and learn from the experience of rewriting all the first batch of components. When we'll be done, we'll have a much better understanding of how to effectively create shared utilities to standardise and accelerate components development even more.

new-components/scanners/bandit/docker-compose.yaml
northdpole: i just copied the example .env, the gosec one also doesn't have any env vars and has the postgresql things

new-components/scanners/bandit/.env
```diff
@@ -0,0 +1,6 @@
+# This is for local setup only.
+SMITHY_INSTANCE_ID=8d719c1c-c569-4078-87b3-4951bd4012ee
+SMITHY_LOG_LEVEL=debug
+SMITHY_IS_LOCAL=false
+SMITHY_STORE_TYPE=postgresql
```
andream16, line 5: You can remove both of this and the following env var so that the component will use SQLite out of the box.

new-components/scanners/bandit/.env
```diff
@@ -0,0 +1,6 @@
...
+SMITHY_REMOTE_STORE_POSTGRES_DSN="postgresql://smithy:smithy1234@findings-db:5432/findings-db?sslmode=disable&connect_timeout=10"
```
andream16, line 6: We should add the env vars that are mentioned in the README here :)

andream16 (approved):

PR #705, opened 2025-01-15
[ocu-472] basic smithyctl boilerplate

andream16:
- Adding boilerplate for `smithyctl` commands. Will open another PR shortly implementing packaging.
- Fixing a small bug on V1 Types parsing. Also moving exported code in `smithyctl/pkg`
- Making sure that we can test every single go test file on CI - some were being skipped and now it's fixed. Opened https://linear.app/smithy/issue/OCU-482/make-sure-git-cloner-integration-test-works-on-ci to tackle an issue that I found on CI with a test. I spent some time on it but it would be better for me to focus on what matters more right now. I have an action to work on this when I have more time.


smithyctl/pkg/types/v1/parameter.go
```diff
@@ -57,7 +57,9 @@ func (p *Parameter) UnmarshalJSON(b []byte) error {
 	case ParameterTypeString, ParameterTypeConststring:
 		parameterValueStrPtr := &struct{ Value *string }{}
 		if err = json.Unmarshal(b, parameterValueStrPtr); err == nil {
-			p.Value = parameterValueStrPtr.Value
+			if parameterValueStrPtr.Value != nil {
```
andream16, line 60: Small fix to an issue that I discovered while re-checking all the tests.

Makefile
```diff
@@ -133,7 +134,10 @@ install-go-test-tools:
 
 go-tests:
```
andream16, line 135: This now covers EVERY possible go test file in the repo. Before there were some issues due to multiple go.mods being present and test runs being hidden on CI.

Now this enable for a sane framework

ptzianos:
A couple of general comments that don't fit neatly into any of the commits:
1. please move the v1 types out of the smithyctl subtree and into the pkg subtree, they are supposed to be used in other places and they are not just for smithyctl.
2. please move the smithyctl commands away from the internal packages and somewhere where they are available from other modules so that we can repackage them

smithyctl/internal/command/packaging/command.go
```diff
@@ -0,0 +1,17 @@
+package packaging
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// NewCommand returns a new package command.
+func NewCommand() *cobra.Command {
+	// TODO: implement in the next PR.
```
dlicheva, line 9: Is there a ticket associated with this todo?

smithyctl/internal/command/version/command.go
```diff
@@ -0,0 +1,15 @@
+package version
+
+import "github.com/spf13/cobra"
+
+// NewCommand returns a new version command.
+func NewCommand() *cobra.Command {
```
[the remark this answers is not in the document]
andream16: If you don't mind let's do it in its ticket

smithyctl/internal/command/packaging/command.go
andream16: Done

andream16:
> A couple of general comments that don't fit neatly into any of the commits:
> 
> 1. please move the v1 types out of the smithyctl subtree and into the pkg subtree, they are supposed to be used in other places and they are not just for smithyctl.
> 2. please move the smithyctl commands away from the internal packages and somewhere where they are available from other modules so that we can repackage them

1. Done
2. Exported commands in PKG for smithyctl

andream16:
Declining in favour of https://github.com/smithy-security/smithy/pull/707

PR #707, opened 2025-01-16
[ocu-472] basic smithyctl boilerplate + small fixes

andream16:
Re-opened from https://github.com/smithy-security/smithy/pull/705

---

- Adding boilerplate for `smithyctl` commands. Exported commands for SAAS are defined in `command`. Will open another PR shortly implementing packaging.
- Fixing a small bug on V1 Types parsing. Also moving exported code in `pkg`
- Making sure that we can test every single go test file on CI - some were being skipped and now it's fixed. Opened https://linear.app/smithy/issue/OCU-482/make-sure-git-cloner-integration-test-works-on-ci to tackle an issue that I found on CI with a test. I spent some time on it but it would be better for me to focus on what matters more right now. I have an action to work on this when I have more time.
- Fixed an issue with the bandit scanner not building correctly due to wrong import
- Refreshed all dependencies
- Make sure that every single test in the repo is executed and passes


pkg/types/v1/parameter.go
```diff
@@ -57,7 +57,9 @@ func (p *Parameter) UnmarshalJSON(b []byte) error {
 	case ParameterTypeString, ParameterTypeConststring:
 		parameterValueStrPtr := &struct{ Value *string }{}
 		if err = json.Unmarshal(b, parameterValueStrPtr); err == nil {
```
andream16, line 59: Small fix to an issue that I discovered while re-checking all the tests.

tools.go
```diff
@@ -0,0 +1,10 @@
+//go:build tools
```
andream16, line 1: This has been moved from `smithyctl/tools.go`

smithyctl/pkg/command/command.go
```diff
@@ -0,0 +1,30 @@
+package command
```
andream16, line 1: Exporting command to be imported by SAAS.

new-components/scanners/bandit/cmd/main.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/smithy-security/smithy/sdk/component"
 
-	"github.com/smithy-security/smithy/new-components/scanners/gosec/internal/transformer"
+	"github.com/smithy-security/smithy/new-components/scanners/bandit/internal/transformer"
```
andream16, line 12: We were importing the wrong package here and failing to build as we can't import internal packages from other modules

Makefile
```diff
@@ -133,7 +134,16 @@ install-go-test-tools:
 
 go-tests:
 	@mkdir -p tests/output
-	@gotestsum --junitfile tests/output/unit-tests.xml -- -race -coverprofile tests/output/cover.out $(GO_TEST_PACKAGES)
+	@for package in $(GO_TEST_PACKAGES); do \
+		if [ "$$package" = "./components/producers/golang-nancy/examples" ]; then \
```
andream16, line 138: We skip this as in this path there's an example vulnerable `go.mod`

ptzianos: can't we use a `-not` flag in the find command above to exclude this?

andream16: Fair, I amended the command to ignore the path :)

smithyctl/internal/command/root.go
```diff
@@ -0,0 +1,19 @@
+package command
```
andream16, line 1: This is internal only as it will be used only to put together the OSS smithyctl binary.

ptzianos (approved):

smithyctl/internal/command/root.go
andream16: Let's keep as is for now if you don't mind - if it's problematic we can refactor it in the next PRs :)

PR #711, opened 2025-01-20
Additional Finding Info protos

PTruscott:
These will be used as annotations for what the additional information the data-enricher returns

### DIFF
--- a/proto/ocsf/ocsf_ext/finding_info/v1/finding_info.proto
+++ b/proto/ocsf/ocsf_ext/finding_info/v1/finding_info.proto
@@ -47,3 +47,29 @@ message DataSource {
     PurlFindingLocationData purl_finding_location_data = 4;
   }
 }
+
+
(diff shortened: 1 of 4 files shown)


proto/ocsf/ocsf_ext/finding_info/v1/finding_info.proto
```diff
@@ -47,3 +47,24 @@ message DataSource {
...
+  }
+
+  message CorrelationEnrichment {
+    repeated string findings = 1;
+  }
+
+  EnrichmentType enrichment_type = 1;
+
+  oneof enrichment {
+    bool duplicate = 2;
```
andream16, line 67: I think that it'd be better to use messages here so that they are open for changes directly

For example, we might want to add some extra fields like num_of_duplicates to the duplication enrichment and so

By having message there's no problem adding such property :)

andream16 (approved):

PR #718, opened 2025-01-22
Changed findings to use the correct datatype

PTruscott:


proto/ocsf/ocsf_ext/finding_info/v1/finding_info.proto
```diff
@@ -57,11 +58,12 @@ message Enrichment {
   }
 
   // if it has this enrichment at all, the finding is a duplicate
-  message DuplicateEnrichment {}
+  message DuplicateEnrichment {
```
andream16, line 61: It looks like this needs formatting

andream16:
The git log seems a bit off - perhaps you could make sure that changes to the same files are in the same commit

andream16 (approved):

PR #722, opened 2025-01-27
ocu-473  - Package  and push components using oras

andream16:
This PR implements `smithyctl component package`. 

**The process is documented in `smithyctl/README.md` and in [notion](https://www.notion.so/use-smithy/Spec-Smithyctl-redesign-17784054488e80ad89b6d6b0a6422746) so please have a look! before reviewing this PR**

In short:
- we look for a `component.yaml` containing the component configuration
- we "package" it - aka get its encoded representation and push it to the OCI registry (local or remote)

---

Changes in this PR:
- JSON Schema for V1 types which are also used for validating component configuration in `component.yaml`
- Packaging logic
- GH Action to package a component on release, matching the correct git tag
- Docs

Changes to be made separately:
- Building and Pushing container images
- Running components locally
- Publishing smithyctl binaries
- Injecting correctly the version into smithyctl binaries at build time

---

**How to test:**

Issue a GitHub PAT with permission enabled to write/read packages

*Local*:
- Run `make smithyctl/cmd/bin`

With `remote` OCI registry run: 
```shell
./bin/smithyctl/cmd/linux/amd64/smithyctl --debug-enabled=true --debug-level=debug component package --registry-auth-enabled=true --registry-auth-username=$USERNAME --registry-auth-password=$PAT --registry-url=ghcr.io --sdk-version=v1.0.0 --spec-path=smithyctl/command/component/testdata/package/valid/component.yaml --version=v3.3.0
```

With `local` registry run:

To spin up the registry
```shell
docker run --publish "127.0.0.1:5000:5000" --name kind-registry registry:2
```

In another terminal, to interact with it
```shell
./bin/smithyctl/cmd/linux/amd64/smithyctl --debug-enabled=true --debug-level=debug component package --registry-base-repository=smithy-security/smithy/manifests --registry-url=127.0.0.1:5000 --sdk-version=v1.0.0 --spec-path=smithyctl/command/component/testdata/package/valid/component.yaml --version=v3.2.1
```

*On CI*:
This will basically run a GitHub Action locally to package the component

- Create an example `component.yaml` in `new-components/scanners/gosec`. You can copy a valid one from `smithyctl/command/testdata/package`
- Install [act](https://github.com/nektos/act)
- Create a new example tag like `git tag scanners/gosec/v0.0.1`
- Run
```shell
act -j package \           
  -e <(echo '{"ref":"refs/tags/scanners/gosec/v1.0.0"}') \
  --actor=$GH_USERNAME \
  --secret GITHUB_TOKEN=$GH_PAT \
  --container-architecture linux/amd64
```
 - Check the uploaded result in https://github.com/orgs/smithy-security/packages/container/package/manifests%2Fcomponents%2Ftarget%2Fexample-component


pkg/types/v1/component.schema.json
```diff
@@ -0,0 +1,85 @@
+{
```
andream16, line 1: Helps us validating input `smithy-component.yaml`

pkg/types/v1/component.go
```diff
@@ -9,30 +16,56 @@ type (
...
-		Script string
+		Script string `json:"script" yaml:"script"`
 	}
 )
+
+//go:embed component.schema.json
+var componentJSONSchema string
+
+// Validate validates a component spec.
+func (c *Component) Validate() error {
```
andream16, line 51: Re-usable method to validate specs

pkg/types/v1/parameter.go
```diff
@@ -113,15 +113,27 @@ func (p *Parameter) Validate() error {
...
+	}
+
 	var correctType bool
+
 	switch p.Type {
 	case ParameterTypeString, ParameterTypeConststring:
-		_, correctType = p.Value.(*string)
-		if !correctType {
-			_, correctType = p.Value.(string)
+		var s string
```
andream16, line 125: Simplifies this and we can avoid going crazy with pointers vs values

smithyctl/internal/command/root.go
```diff
@@ -2,15 +2,50 @@ package command
...
+	debugLevel   string
+}
+
 // NewRootCommand register all the subcommands into the same root command.
 func NewRootCommand(commands ...*cobra.Command) *cobra.Command {
 	rootCmd := &cobra.Command{
 		Use:   "smithyctl",
 		Short: "A CLI app for managing components and workflows",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if flags.debugEnabled {
```
andream16, line 22: Global logger available from main command and initialised by global flags

smithyctl/internal/registry/registry.go
```diff
@@ -0,0 +1,191 @@
...
+		oras.PackManifestOptions{
+			Layers: []ocispec.Descriptor{blobDescriptor},
+			ManifestAnnotations: map[string]string{
+				"smithy.sdk.version":           sdkVersion,
+				"smithy.component.description": component.Description,
+				"smithy.component.name":        component.Name,
+				"smithy.component.type":        component.Type.String(),
+				"smithy.component.version":     componentVersion,
+				"smithy.component.source": fmt.Sprintf(
+					"new-components/%s/%s",
```
andream16, line 161: This is temporary and needs to be improved a bit

smithyctl/internal/registry/registry.go
```diff
@@ -0,0 +1,191 @@
...
+	reg, err := remote.NewRegistry(registryHost)
+	if err != nil {
+		return nil, errors.Errorf("could not create registry for host '%s': %w", registryHost, err)
+	}
+
+	var regAuthClient = &auth.Client{
+		Cache: auth.NewCache(),
+	}
+
+	if registryAuthEnabled {
```
andream16, line 65: This bit changes based on running locally vs remotely

smithyctl/internal/registry/registry.go
```diff
@@ -0,0 +1,191 @@
...
+		regAuthClient.Credential = auth.StaticCredential(
+			registryHost,
+			auth.Credential{
+				Username: registryAuthUsername,
+				Password: registryAuthPassword,
+			},
+		)
+		regAuthClient.Client = retry.DefaultClient
+	} else {
+		reg.PlainHTTP = true
```
andream16, line 75: Necessary if auth is off

smithyctl/internal/registry/registry.go
```diff
@@ -0,0 +1,191 @@
...
+	var regAuthClient = &auth.Client{
+		Cache: auth.NewCache(),
+	}
+
+	if registryAuthEnabled {
+		regAuthClient.Credential = auth.StaticCredential(
+			registryHost,
+			auth.Credential{
+				Username: registryAuthUsername,
+				Password: registryAuthPassword,
```
andream16, line 70: This is VERY confusing but can actually be a PAT

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
+name: Package on Tag
+
+on:
+  push:
+    tags:
+      - '*/*/*/v*.*.*'
```
andream16, line 6: Triggered on tags like `scanners/gosec/v1.0.0`

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
...
+          VERSION=$(echo $TAG | cut -d/ -f3)
+          
+          echo "component_type=$COMPONENT_TYPE" >> $GITHUB_OUTPUT
+          echo "component_name=$COMPONENT_NAME" >> $GITHUB_OUTPUT
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+      - name: Extract SDK Version
+        id: sdk_version
+        run: |
+          SDK_VERSION=$(grep 'github.com/smithy-security/smithy/sdk' new-components/${{ steps.tag_info.outputs.component_type }}/${{ steps.tag_info.outputs.component_name }}/go.mod | awk '{print $2}')
```
andream16, line 29: Dinamically extract SDK version from deps

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
...
+          echo "sdk_version=$SDK_VERSION" >> $GITHUB_OUTPUT
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
+
+      # To be replaced with compiled and release binary - TODO: https://linear.app/smithy/issue/OCU-480/build-and-push-binaries
```
andream16, line 37: As per comment - when we publish smithyctl binaries we can re-use them here since they'll be versioned

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Extract SDK Version
+        id: sdk_version
+        run: |
+          SDK_VERSION=$(grep 'github.com/smithy-security/smithy/sdk' new-components/${{ steps.tag_info.outputs.component_type }}/${{ steps.tag_info.outputs.component_name }}/go.mod | awk '{print $2}')
+          echo "sdk_version=$SDK_VERSION" >> $GITHUB_OUTPUT
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
```
dlicheva, line 35: Go 1.22 has been released. Should we use that instead?

pkg/types/v1/component.schema.json
```diff
@@ -0,0 +1,85 @@
...
+            "pattern": "^(?:[a-zA-Z0-9._-]+(?:/[a-zA-Z0-9._-]+)*)(?::[a-zA-Z0-9._-]+|@sha256:[a-fA-F0-9]{64})?$"
+          },
+          "script": {
+            "type": "string"
+          }
+        }
+      }
+    }
+  },
+  "additionalProperties": false
```
dlicheva, line 84: What if `additionalProperties` is true? What other properties are available?

smithyctl/README.md
```diff
@@ -0,0 +1,77 @@
...
+  component \
+    package \
+        --registry-auth-enabled=true \
+        --registry-auth-username=smithy \
+        --registry-auth-password=XXX \
+        --registry-url=ghcr.io \
+        --sdk-version=v1.0.0
+        --spec-path=new-components/scanners/bandit
+        --version=v3.2.1
+```
```
dlicheva, line 77: I think it would be nice to add instructions on how to build and publish a component, similar to how you've described them in the PR.

dlicheva:
Nice work @andream16 ! The publishing works locally and on Git. Is it possible to delete a published component from the registry? I pushed this as a test: https://github.com/orgs/smithy-security/packages/container/package/manifests%2Fcomponents%2Ftarget%2Fgosec-test

pkg/types/v1/component.schema.json
andream16: None, this basically would make the validation fail if more fields are provided in the object :)

smithyctl/command/component/package.go
```diff
@@ -0,0 +1,208 @@
...
+	"os"
+	"path/filepath"
+	"slices"
+
+	"github.com/go-errors/errors"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	v1 "github.com/smithy-security/smithy/pkg/types/v1"
+
```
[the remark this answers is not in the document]
andream16: I agree but the linter insists it's correct like this when I reformat XD I think we need to give the fmt a bit more love

smithyctl/command/component/package.go
```diff
@@ -0,0 +1,208 @@
...
+		ComponentVersion: flags.packageVersion,
+	}); err != nil {
+		return errors.Errorf("failed to package component: %w", err)
+	}
+
+	return nil
+}
+
+func parseComponentSpec(path string) (*v1.Component, error) {
+	const (
```
ptzianos, line 147: this looks like it could have side effects by forgotten package files in the root of the repo, I would rather we keep it simple and check if the path exists, is a file and if it can be parsed. in any other case we just return a clear error and the user should figure out what they did wrong

smithyctl/command/component/package.go
```diff
@@ -0,0 +1,208 @@
...
+	info, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, errors.New(fmt.Sprintf("%s does not exist", path))
+		}
+		return nil, fmt.Errorf("failed check config file: %w", err)
+	}
+
+	var fullPath = path
+	// If the path is a directory we check if it contains a smithy-component.yaml or smithy-component.yml.
```
ptzianos, line 167: I don't see why we need to call them `smithy-component.yaml`. It's a smithy package so we can just expect a component.yaml, same as kustomize for example that expects a kustomization.yaml. Anything else doesn't work and we show an error

smithyctl/command/component/package.go
```diff
@@ -0,0 +1,208 @@
...
+	if info.IsDir() {
+		files, err := os.ReadDir(path)
```
ptzianos, line 169: this could be simplified to `is this a directory? if yes append component.yaml to it and call myself in a recursion for the new path`

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
...
+jobs:
+  package:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Extract tag details
+        id: tag_info
+        run: |
+          TAG=${GITHUB_REF#refs/tags/}
```
ptzianos, line 17: at this point we should also do a check to make sure that the tag refers to an actual component

.github/workflows/package-component.yaml
ptzianos, line 29: can we move this into a make target so that we can also run it locally and get the same result?

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Extract SDK Version
+        id: sdk_version
+        run: |
+          SDK_VERSION=$(grep 'github.com/smithy-security/smithy/sdk' new-components/${{ steps.tag_info.outputs.component_type }}/${{ steps.tag_info.outputs.component_name }}/go.mod | awk '{print $2}')
+          echo "sdk_version=$SDK_VERSION" >> $GITHUB_OUTPUT
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.23'
```
ptzianos, line 35: this should use the go.mod file version

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+      # To be replaced with compiled and release binary - TODO: https://linear.app/smithy/issue/OCU-480/build-and-push-binaries
+      - name: Build CLI
+        working-directory: ./smithyctl
+        run: go build -o ../cli ./cmd/main.go
```
ptzianos, line 40: you can run `make cmd/smithyctl/bin` here

.github/workflows/package-component.yaml
```diff
@@ -0,0 +1,58 @@
...
+          USERNAME: ${{ secrets.REGISTRY_USERNAME }}
+          PAT: ${{ secrets.REGISTRY_PAT }}
+        run: |
+          ./cli \
+            --debug-enabled=true \
+            --debug-level=debug \
+            component \
+              package \
+                --registry-auth-enabled=true \
+                --registry-auth-username=${{ env.USERNAME }} \
```
ptzianos, line 53: we shouldn't be using username and PAT because they are going to go stale and cause issues eventually. since we are using ghcr.io we can use the same method as the current publish workflow which logs in with Docker and Helm is able to pick up the same credentials are re-use them and those credentials are the workflows own credentials: https://github.com/smithy-security/smithy/blob/main/.github/workflows/publish.yml#L21. no extra stuff needed

smithyctl/command/component/package.go
```diff
@@ -0,0 +1,208 @@
...
+	cmd.
+		Flags().
+		StringVar(
+			&flags.packageVersion,
+			"version",
+			"",
+			"the version to be used to package the component",
+		)
+	_ = cmd.MarkFlagRequired("version")
+	cmd.
```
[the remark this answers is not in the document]
andream16: This is meant to be the package version - so we're versioning the actual package itself

The binary can be separately versioned :)

I agree that we need to inject the binary version at build time - but that's not done in this PR

smithyctl/command/component/package.go
andream16: My idea is that in general it's nice to have this at root level and just rely on default behaviour

This is often the case with lots of tools - like wrangler etc

I think that we can pass a full path, potentially - but if you feel strongly I can change it - I don't mind

I do think that the proposal is better in terms of UX though!

.github/workflows/package-component.yaml
andream16: I think this is going to be replaced real soon with a pre-compiled and versioned binary - so we can avoid breaking the action while merging new unreleased things in to smithyctl

.github/workflows/package-component.yaml
andream16: Done

.github/workflows/package-component.yaml
andream16: Done

.github/workflows/package-component.yaml
andream16: Done by being stricter with the tag match

smithyctl/command/component/package.go
andream16: I ended up using the file directly

smithyctl/command/component/package.go
andream16: Done

smithyctl/command/component/package.go
andream16: I ended up using the file directly for simplicity

pkg/types/v1/parameter.go
andream16: Fixed

pkg/types/v1/parameter.go
```diff
@@ -122,6 +123,19 @@ func (p *Parameter) Validate() error {
 		}
 	case ParameterTypeListstring:
 		_, correctType = p.Value.([]string)
+		if !correctType {
```
andream16, line 126: Covers also from dealing with `[]interface{}` as it happens when we unmarshal `component.yaml` into a Parameter actually 🦀

ptzianos: how would a YAML fragment look that would trigger this? I can see a situation with JSON where this could happen but it would mean that the user has added numbers in the list but in YAML it looks like everything is considered a string: https://go.dev/play/p/btjnSip-iKk

also, how would it work that a list is of type []any but it has strings in it? and I don't via a test where you can force the type, but can you replicate this situation with a YAML or a JSON marshaler?

smithyctl/command/component/package.go
```diff
@@ -0,0 +1,190 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/go-errors/errors"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
```
ptzianos, line 13: nit: extra empty line here

smithyctl/command/component/package.go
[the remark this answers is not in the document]
ptzianos: can you please create a ticket to track injecting version numbers into the binary?

.github/workflows/package-component.yaml
ptzianos: what if we accidentally push a change that breaks smithyctl packaging itself? then the good smithyctl version builds the broken one and pushes it to the registry which causes followup publish steps to fail.

we should build smithyctl itself from the current commit and use that to push itself to the registry. if the command is broken in some way, the next commit will fix it and the workflow will finish successfully without manual intervention.

we are using the go.mod file to set the version automatically in the test workflow: https://github.com/smithy-security/smithy/blob/main/.github/workflows/test.yml#L34

smithyctl/command/component/package.go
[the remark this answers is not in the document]
andream16: Added details here https://linear.app/smithy/issue/OCU-471/build-components

.github/workflows/package-component.yaml
andream16: Smithyctl is not a component, so it is packaged and distributed differently.

I would prefer to use it pre-compiled, tested and versioned after release to keep things stabler and less prone to break accidentally. 

That said, I'll make the changes and can revisit later if needed

pkg/types/v1/parameter.go
andream16: Took this offline - ended up implementing a custom YAML marshaler/unmarshaler like we did for JSON for parameters. Tested and now works as expected even with a component.yaml

smithyctl/command/component/package.go
andream16: `make fmt` keeps adding the line 🤔 

Do you mind if we ignore this for now?

I think we need better support for `go imports` regardless

.github/workflows/package-component.yaml
andream16: Done

pkg/types/v1/parameter.go
```diff
@@ -30,14 +32,81 @@ type (
...
+		return errors.Errorf("failed to unmarshal parameter %s: %v", value.Value, err)
+	}
+
+	p.Name = partialParameter.Name
+	p.Type = partialParameter.Type
+
+	switch partialParameter.Type {
+	case ParameterTypeString, ParameterTypeConststring:
+		parameterValueStrPtr := &struct{ Value *string }{}
+		if err = value.Decode(&parameterValueStrPtr); err == nil {
```
ptzianos, line 87: I think it would make sense to set the parameter value to whatever is the value unmarshalled. if the value is a nil pointer, it's still type checked a nil string pointer that we can check when resolving the parameter overrides

andream16: Can you be clearer? This is mimicking what we do for the JSON unmarshaler as well

My understanding is that both pointers and values can be supplied with `ParameterTypeString` and `ParameterTypeConststring` and this takes care of it - you can check-out the tests in `parameter_test.go`

This also justifies the comment on line :93

ptzianos: as far as I understand the behaviour of the YAML unmarshaller, there are 3 cases that we need to be able to handle:
1. empty string
```yaml
value: ''
```

this is different than empty value, it is a literal empty string and the YAML unrmashaler would return a non-nil pointer pointing to an empty string, right? in which case you would get a nil err out of this line: `err = value.Decode(&parameterValueStrPtr)` and that branch would be executed

2. non-empty string -> the easiest one
again the you get a nil err out of that line of code and that branch gets executed

4. value is not present which is interpreted as a value that needs to be set explicitly by the user, otherwise the workflow launch should return an error. in this case the parameter would look like this:
```yaml
name: some-parameter
type: string
```

in this case the only way for us to figure out that the parameter is missing is to have nil string pointer and the YAML unmarshaler parsing this would set the value of the parameterValueStrPtr to nil, right? which again would mean that the following piece of code would not be executed

andream16: Took this one offline - the TL;DR is that we want to deal with potentially having nil values - by default we use `*string` and that's about it

pkg/types/v1/parameter.go
```diff
@@ -30,14 +32,81 @@ type (
...
+			if parameterValueStrPtr.Value != nil {
+				p.Value = *parameterValueStrPtr.Value
+			}
+			break
+		}
+		parameterValueStr := &struct{ Value string }{}
```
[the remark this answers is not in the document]
andream16: Fixed in previous comment.

ptzianos (approved):

ptzianos (approved):

PR #727, opened 2025-02-01
Slack to new sdk

northdpole:


new-components/reporters/slack/cmd/main.go
```diff
@@ -0,0 +1,39 @@
+package main
+
+import (
+	"context"
+	"log"
+	"time"
+
+	"github.com/go-errors/errors"
```
andream16, line 8: ⛏️ needs formatting

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+	SlackWebhook string
+}
+
+type slackLogger struct {
+	Conf *Conf
+}
+
+// NewSlackLogger returns a new slack logger.
+func NewSlackLogger(c *Conf) slackLogger {
+	return slackLogger{
```
andream16, line 31: Worth checking for conf to not be nil here

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+
+	webhook, err := env.GetOrDefault(
+		"SLACK_WEBHOOK",
+		"",
+		append(envOpts, env.WithDefaultOnError(false))...,
+	)
+	if err != nil {
+		return nil, errors.Errorf("could not get env variable for SLACK_WEBHOOK: %w", err)
+	}
+	if webhook == "" {
```
andream16, line 51: I think this is not necessary as `env.WithDefaultOnError` is set to false

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return nil
+}
+
+func (s slackLogger) push(b string, webhook string) error {
+	type SlackRequestBody struct {
+		Text string `json:"text"`
+	}
+	var err error
+	body, _ := json.Marshal(SlackRequestBody{Text: b})
```
andream16, line 89: Let's check for the error here plz

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+		//  TODO: when duplicate enrichment is available, add: out of which, <newResults> new", "the template to use when posting the results"
+		return s.push(msg, s.Conf.SlackWebhook)
+	}
+	logger.Error("Received 0 scans, this is likely an error, skipping sending empty message")
+
+	return nil
+}
+
+func (s slackLogger) push(b string, webhook string) error {
+	type SlackRequestBody struct {
```
andream16, line 85: Can we move this with the other types for consistency?

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+		Text string `json:"text"`
+	}
+	var err error
+	body, _ := json.Marshal(SlackRequestBody{Text: b})
+	req, err := http.NewRequest(http.MethodPost, webhook, bytes.NewBuffer(body))
```
andream16, line 90: Let's use `http.NewRequestWithContext` and pass the report CTX in input here :)

So we can acknowledge cancellations and in future tracing

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	buf := new(bytes.Buffer)
+	if _, err := buf.ReadFrom(resp.Body); err != nil {
```
andream16, line 102: Should we check for the status code here?

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+		return fmt.Errorf("could not read from resp: %w", err)
+	}
+	if strings.ToLower(buf.String()) != "ok" {
+		return errors.New("non-ok response returned from Slack")
+	}
+	return nil
+}
+
+// func (s slackLogger) countNewFindings(findings []*vf.VulnerabilityFinding) int {
```
andream16, line 111: It looks like this can be removed

new-components/reporters/slack/internal/reporter/reporter.go
northdpole: not sure what you mean here

andream16: I mean that the check for the empty webhook is superfluous because `env.WithDefaultOnError(false)` already returns an error when the value is empty :)

andream16 (approved):

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+
+	SlackRequestBody struct {
+		Text string `json:"text"`
+	}
+)
+
+// NewSlackLogger returns a new slack logger.
+func NewSlackLogger(c *Conf) (*slackLogger, error) {
+	if c == nil {
+		return nil, errors.Errorf("configuration is nil")
```
andream16, line 38: Can be replaced with `errors.New` as we're not formatting :)

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+// Report logs the findings summary in slack.
+func (s slackLogger) Report(
+	ctx context.Context,
+	findings []*vf.VulnerabilityFinding,
+) error {
+	logger := component.
+		LoggerFromContext(ctx).
+		With(slog.Int("num_findings", len(findings)))
+
+	if len(findings) > 0 {
```
andream16, line 75: We can reverse this and return earlier for no findings

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+		scanID := findings[0].Finding.FindingInfo.Uid
```
andream16, line 76: How come we don't do it for all findings? I guess we could handle that with a for loop

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, errors.Errorf("could not get env variable for SLACK_WEBHOOK: %w", err)
+	}
+
+	return &Conf{
+		SlackWebhook: webhook,
+	}, nil
+}
+
+// Report logs the findings summary in slack.
+func (s slackLogger) Report(
```
andream16, line 67: It would help adding a few more debug logs!

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (s slackLogger) push(ctx context.Context, b string, webhook string) error {
+
+	var err error
+	body, err := json.Marshal(SlackRequestBody{Text: b})
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, webhook, bytes.NewBuffer(body))
```
andream16, line 97: Can we leverage a context with timeout here to make sure that we're not stuck forever? 10 seconds should do

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
andream16, line 104: Missing `defer resp.Body.Close()` 👀

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,146 @@
...
+		return errors.Errorf("failed to submit to slack, status code: %d", resp.StatusCode)
+	}
+
+	buf := new(bytes.Buffer)
+	if _, err := buf.ReadFrom(resp.Body); err != nil {
+		return fmt.Errorf("could not read from resp: %w", err)
+	}
+
+	if strings.ToLower(buf.String()) != "ok" {
+		return errors.New("non-ok response returned from Slack")
```
andream16, line 119: Worth logging the response if invalid imo

new-components/reporters/slack/internal/reporter/reporter.go
northdpole: my assumption was that this is called for a single scan so everything should have the same scanID

new-components/reporters/slack/internal/reporter/reporter.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	ctx, cancelFunc := context.WithTimeout(ctx, time.Duration(10*time.Second))
+	defer cancelFunc()
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, webhook, bytes.NewBuffer(body))
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	client := &http.Client{}
```
andream16, line 109: It would nicer to create the client before the execution in the main and inject it into `slackLogger` - it would also enable for better testing as we could mock `Do`

ptzianos:
merged as part of https://github.com/smithy-security/smithy/pull/813

PR #754, opened 2025-02-25
fix typo in parameter type names

ptzianos:


andream16:
How did you generate the enums to fix the case?

Can you put the fix in `tools.go` please?

andream16:
> > How did you generate the enums to fix the case?
> > Can you put the fix in `tools.go` please?
> 
> it was the easiest method ever to fix them.... manually

Perhaps there's a way to fix this with the enum gen?

ptzianos:
@andream16 figured out how to do this based on the features of the library without manual stuff or hacks

andream16 (approved):

PR #759, opened 2025-03-08
smithyctl: add dynamic image resolution and building when running workflows

ptzianos:


northdpole (approved):

northdpole (approved):

Makefile
```diff
@@ -103,6 +103,11 @@ clean-migrations-compose:
 
 clean: clean-protos clean-migrations-compose
 
+smithyctl/bin:
```
andream16, line 106: I think we have this already defined as `smithyctl/cmd/bin` - you might want to remove that

smithyctl/internal/images/component.go
```diff
@@ -0,0 +1,193 @@
...
+// images of components managed by the system
+type resolutionOptions struct {
+	registry  string
+	namespace string
+	tag       string
+}
+
+// ResolutionOptionFn is used to define common attributes of all the images
+// of components
+type ResolutionOptionFn func(*resolutionOptions)
```
andream16, line 50: Nit but it would be nice to return an erro if the passed arg is misconfigured :)

So `type ResolutionOptionFn func(*resolutionOptions) error`

smithyctl/internal/images/docker/builder.go
```diff
@@ -0,0 +1,253 @@
...
+// images that will be built for the Builder
+type builderOptions struct {
+	platform           string
+	baseDockerfilePath string
+	push               bool
+	labels             map[string]string
+}
+
+// BuilderOptionFn is used to modify the options passed to the builder
+type BuilderOptionFn func(*builderOptions)
```
andream16, line 32: Same comment here

andream16:
Can we share the same Docker client in `internal > engine > docker > client`?

andream16 (approved):

andream16 (approved):

Makefile
ptzianos: that one you mention is for the old smithyctl binary

PR #769, opened 2025-03-14
Replacing Tekton world docs with Smithyctl/SDK world + small tweaks for trying out smithy

andream16:
- Removing legacy docs, targets and deployments
- Removing embedded platform used to build containers for new components as this fails on Darwin
- Fixing a bug that removed the chance for components to share the same local database
- Fixing a bug that did not allow to install smithyctl from source
- Using docker client negotiation also when running components
- Adding new documentation
- Refreshing deps

---

How to test:
- go inside `smithyctl` and run `go install`. `smithyctl` should now be installed in your system.
- go back to root
- run `smithyctl workflow run --spec-path=examples/golang/workflow.yaml --build-component-images=true` and watch a workflow succeed.

---

After this is merged:
- I'll release smithyctl and try to install latest from source


northdpole (approved):

dlicheva (approved):
Works wonderfully!

PR #770, opened 2025-03-15
bandit example workflow and component.yaml

northdpole:


new-components/scanners/bandit/cmd/main.go
```diff
@@ -16,6 +16,7 @@ func main() {
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
 	defer cancel()
 
+	log.Println("running bandit")
```
andream16, line 19: I don't think that we need these logs

new-components/scanners/bandit/internal/transformer/transformer.go
```diff
@@ -135,7 +135,7 @@ func New(opts ...BanditTransformerOption) (*BanditTransformer, error) {
 	rawOutFilePath, err := env.GetOrDefault(
 		"BANDIT_RAW_OUT_FILE_PATH",
 		"bandit.json",
-		env.WithDefaultOnError(true),
+		env.WithDefaultOnError(false),
```
andream16, line 138: Are we sure we're not ok with having the default if this is missing? 🤔

northdpole: we really want this to fail if there is no input bandit file

andream16 (approved):

PR #772, opened 2025-03-17
Use default temp directory instead of hardcoded `/tmp`

nscuro:
This is mostly relevant for component packaging, as `/tmp` does not exist on Windows, preventing the poor souls having to work with that OS from running `smithyctl component package`.

Made the same change in other locations for consistency.


andream16:
Thanks for the PR @nscuro !

We're going to deprecate these components real soon so this won't interfere with your new OSS SDK experience - do you mind if we close this for now?

nscuro:
@andream16 Absolutely no problem, please feel free to close.

PR #773, opened 2025-03-17
Use `filepath` over `path` for path joining and comparison on `localFileLoader`

nscuro:
The `filepath` package is more OS-agnostic, allowing the local file loader to work on Windows.


andream16 (commented):
Thanks for the addition - lg2m but I'll test this first :)

andream16:
Works well - thanks for the addition!

andream16 (approved):

PR #774, opened 2025-03-17
Add debug log for to-be-applied k8s objects

nscuro:
Allows for easier debugging when applying of objects fails, e.g. because the k8s API server rejects them for unknown reasons.

Wraps the object in a struct to allow for lazy evaluation as per https://pkg.go.dev/log/slog#hdr-Performance_considerations.


pkg/k8s/clientset.go
```diff
@@ -89,6 +94,8 @@ func (c clientSet) Apply(ctx context.Context, obj runtime.Object, namespace stri
 		return errors.Errorf("%s/%s: could not convert runtime object into unstructured data: %w", namespace, name, err)
 	}
 
+	slog.Debug("applying object", "namespace", namespace, "forceConflicts", forceConflicts, "object", lazyObjectLogValue{obj})
```
nscuro, line 97: I just used `slog` here without having a broader understanding how Smithy propagates desired log levels. Happy to make changes so it aligns better with how Smithy approaches logging.

andream16: I think that we might move this internally and remove it from the OSS this week - do you mind if we hold the change for now?

nscuro: Sure, no rush.

PR #786, opened 2025-03-19
refactor codeql runners to not overwrite each other's images

ptzianos:


ptzianos:
no longer relevant, will fix in a different way

PR #796, opened 2025-03-20
Image target & trivy improvements to support both local and remote images

northdpole:


ptzianos:
merged as part of https://github.com/smithy-security/smithy/pull/813

PR #819, opened 2025-03-31
Status investigation protos

northdpole:


Makefile
```diff
@@ -67,7 +67,7 @@ go-cover: go-tests
 tests: test-go
 
 install-go-fmt-tools:
-	@go install github.com/bufbuild/buf/cmd/buf@v1.28.1
+	@go install github.com/bufbuild/buf/cmd/buf@v1.45.0
```
andream16, line 70: Make sure we do the same in the buf Dockerfile

andream16 (approved):

Makefile
northdpole: they are the same now, dockerfile is on 45

PR #831, opened 2025-04-02
Fix pip audit experience

northdpole:


components/scanners/osv-scanner/README.md
```diff
@@ -13,8 +13,8 @@ as the following:
 
 | Environment Variable     | Type   | Required | Default    | Description                                             |
 |--------------------------|--------|----------|------------|---------------------------------------------------------|
-| PIP\_AUDIT\_RAW\_OUT\_FILE\_PATH  | string | yes      | -          | The path where to find the gosec report                 |
-| PIP\_AUDIT\_TARGET\_TYPE         | string | false    | repository | The type of target that was used to generate the report |
+| RAW\_OUT\_FILE\_PATH  | string | yes      | -          | The path where to find the gosec report                 |
```
ptzianos, line 16: the gosec report?

components/scanners/osv-scanner/component.yaml
```diff
@@ -0,0 +1,22 @@
+name: osv-scanner
+description: "Scans projects with google's osv-scanner, parses sarif findings into OCSF format"
+type: scanner
+steps:
+  - name: scanner
+    image: components/scanners/osv-scanner/scanner
+    executable: /entrypoint.sh
```
ptzianos, line 7: this will not work, the executable should be `/bin/bash` and then the args should be
```
- -c
- /entrpoint.sh
- scan source -r --format=sarif --call-analysis=true {{ sourceCodeWorkspace }}
```

components/scanners/osv-scanner/scanner/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
```
ptzianos, line 1: why is this needed? can't we invoke the osv scanner without having an entrypoint?

ptzianos (approved):

components/scanners/osv-scanner/component.yaml
northdpole: are you saying, this won't work on smithy-pro or just the oss?
for reference it works on the oss

andream16 (approved):

PR #842, opened 2025-04-07
wrap trufflehog and add 'write to file' entrypoint'

northdpole:


components/scanners/trufflehog/component.yaml
```diff
@@ -6,6 +6,16 @@ parameters:
     type: "string"
     value: "TARGET_TYPE_REPOSITORY"
 steps:
+  - name: "run-trufflehog"
+    image:  "components/scanners/trufflehog/scanner"
+    executable: /smithy_entrypoint.sh
```
ptzianos, line 11: this looks like it could be simplified by using the trufflehog image directly and changing the executable and args to be like the following:
```yaml
executable: /bin/bash
args:
- -c
- echo "Running trufflehog and sending output to ${RAW_OUT_FILE}"  && /etc/entrypoint.sh filesystem --json --no-fail --directory="{{sourceCodeWorkspace}}  | tee ${RAW_OUT_FILE}"
```

andream16 (approved):

ptzianos (approved):

PR #865, opened 2025-04-14
allow targets to propagate metadata (Fixes OCU-532)

ptzianos:
this PR adds the following things to the OSS tooling and components:
1. it extends the git-clone component to keep track of the metadata of the repository that is cloned
2. it extends gosec and nancy to use these metadata when they are available
3. it extends the smithyctl executor to inject the necessary environment variables and volumes to targets and scanners so that the metadata can be communicated  between the components


andream16:
I remember that we manually made a change in git-clone's vendors, do you mind make sure that's still in place?

andream16 (approved):

PR #876, opened 2025-04-17
Add metadata publishing to image get

northdpole:


components/targets/image-get/component.yaml.orig
```diff
@@ -0,0 +1,30 @@
+name: image-get
```
andream16, line 1: It looks like this should be removed as it's a backup file

components/targets/image-get/metadata-writer/cmd/main.go
```diff
@@ -0,0 +1,193 @@
...
+	return &Conf{
+		TargetMetadataPath: targetMetadataPath,
+		Image:              image,
+	}, nil
+}
+
+func WriteTargetMetadata(conf Conf) error {
+	purl, err := packageUrlFromImage(conf.Image)
+	if err != nil {
+		return errors.Errorf("could not get package url from image: %s", err)
```
andream16, line 65: Should use `%w` and not `%s`

components/targets/image-get/metadata-writer/cmd/main.go
```diff
@@ -0,0 +1,193 @@
...
+
+	marshaledDataSource, err := protojson.Marshal(dataSource)
+	if err != nil {
+		return errors.Errorf("could not marshal data source into JSON: %w", err)
+	}
+
+	// Write content to the file
+	err = os.WriteFile(conf.TargetMetadataPath, marshaledDataSource, 0644)
+	if err != nil {
+		return errors.Errorf("Error writing file: %w\n", err)
```
andream16, line 84: You don't need the `\n` here

components/targets/image-get/metadata-writer/cmd/main.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return nil
+}
+
+func packageUrlFromImage(image string) (*packageurl.PackageURL, error) {
+	namespace := ""
```
andream16, line 90: Could we init these in the same var block?

```go
var (
    path string
    ...
)
```

components/targets/image-get/metadata-writer/cmd/main.go
```diff
@@ -0,0 +1,193 @@
...
+		conf: conf,
+	}, nil
+}
+
+func (t *imageMetadataWriterTarget) Prepare(ctx context.Context) error {
+	if t.conf == nil {
+		return errors.New("conf cannot be nil")
+	}
+
+	if err := WriteTargetMetadata(*t.conf); err != nil {
```
andream16, line 188: Perhaps `WriteTargetMetadata` can accept a pointer so it looks less funky here

components/targets/image-get/skopeo-getter/entrypoint.sh
```diff
@@ -10,7 +10,7 @@ check_var() {
         echo $res
     fi
   else
-    echo "''"
```
andream16, line 13: Is this intentional?

andream16 (approved):

PR #894, opened 2025-04-28
SMITHY-542 add datasource types for website and make zap propagate them

northdpole:


components/scanners/zaproxy/overwrite-metadata/cmd/main.go
```diff
@@ -0,0 +1,59 @@
...
+	"strings"
+
+	"github.com/go-errors/errors"
+	"google.golang.org/protobuf/encoding/protojson"
+
+	"github.com/smithy-security/smithy/sdk/component"
+	ocsffindinginfo "github.com/smithy-security/smithy/sdk/gen/ocsf_ext/finding_info/v1"
+)
+
+func main() {
```
andream16, line 17: Any reason why this is not using the SDK?

components/scanners/zaproxy/go.mod
```diff
@@ -6,10 +6,10 @@ require (
 	github.com/go-errors/errors v1.5.1
 	github.com/jonboulle/clockwork v0.5.0
 	github.com/smithy-security/pkg/env v0.0.1
-	github.com/smithy-security/pkg/sarif v0.0.4
-	github.com/smithy-security/smithy/sdk v0.0.11-alpha
+	github.com/smithy-security/pkg/sarif v0.0.5-0.20250429185943-362b30bf2d13
+	github.com/smithy-security/smithy/sdk v0.0.11-alpha.0.20250429114645-86a9857fcaf9
```
andream16, line 10: If we made any substantial changes to the SDK, can you please bump the actual version?

components/scanners/zaproxy/overwrite-metadata/cmd/main.go
northdpole: can a component have two pieces? this is not necessarily a component, instead this is just a quick hack to propagate metadata for a workflow that doesn't fit squarely into the smithy mold

components/scanners/zaproxy/go.mod
northdpole: yes, please review

-  https://github.com/smithy-security/smithy/pull/894
- https://github.com/smithy-security/pkg/pull/11

then we can merge it and replace the hashes with a version

ptzianos:
@northdpole please fix the dependency versions, the conflicts and cleanup the commit history and this is ready to be merged

components/scanners/zaproxy/go.mod
ptzianos: fixed

components/scanners/zaproxy/overwrite-metadata/cmd/main.go
ptzianos: I think it's an ok solution, the proper long term solution involves a more flexible execution engine and a different type of event as an entrypoint for the state machine

PR #929, opened 2025-05-09
Output git diff file with diff of branch or pr against the base branch

andream16:
Changes:
- We now clone starting from the base branch so we can dynamically resolve it in case an input branch is not the base one. This is usually for anything triggers related or on custom branches.
- We store the base branch and proceed
- We checkout the target branch and do the usual cloning flow
- If the target branch is not the base branch, we output the git diff between them in a `raw.diff` file in the `{{ targetMetadata }}`

I also refactored a bit the target as the responsibilities for git/fs operations were mixed up

---

Tested with this repo and branch!
- manual: `components/targets/git-clone/cmd/test` by using `https://github.com/smithy-security/smithy.git` as ref url and `andream16/feature/smithy-573-output-gitdiff-file-with-diff-of-branch-or-pr-against-the` as reference
- smithyctl:
    1. Update overrides for `examples/golang` to point to the same params as above
    2. Run `smithyctl workflow run --build-component-images=true --overrides=examples/golang/overrides.yaml examples/golang/workflow.yaml`
- Tested also using the default base branch as target and the diff step is skipped as expected :). You can do the same with the step above but not customising the overrides file.

There's a bug with nancy on monorepos so you might want to remove that from the example workflow if you use smithy as target


northdpole:
> Changes:
> 
> * We now clone starting from the base branch so we can dynamically resolve it in case an input branch is not the base one. This is usually for anything triggers related or on custom branches.
> * We store the base branch and proceed
> * We checkout the target branch and do the usual cloning flow
> * If the target branch is not the base branch, we output the git diff between them in a `raw.diff` file in the `{{ targetMetadata }}`
> 
> I also refactored a bit the target as the responsibilities for git/fs operations were mixed up
> 
> Tested with this repo and branch!
> 
> * manual: `components/targets/git-clone/cmd/test` by using `https://github.com/smithy-security/smithy.git` as ref url and `andream16/feature/smithy-573-output-gitdiff-file-with-diff-of-branch-or-pr-against-the` as reference
> * smithyctl:
>   
>   1. Update overrides for `examples/golang` to point to the same params as above
>   2. Run `smithyctl workflow run --build-component-images=true --overrides=examples/golang/overrides.yaml examples/golang/workflow.yaml`
> * Tested also using the default base branch as target and the diff step is skipped as expected :). You can do the same with the step above but not customising the overrides file.
> 
> There's a bug with nancy on monorepos so you might want to remove that from the example workflow if you use smithy as target

this is awesome stuff and it's moving towards a fantastic direction!
edge case: 
what if a pr is not against the main branch but instead against something else say "next"?
wouldn't it be simpler to make the "base" configurable?

northdpole (approved):

PR #997, opened 2025-06-04
git-clone: split away auth token and username into own parameters

ptzianos:


northdpole (approved):

components/targets/git-clone/pkg/git/conf.go
```diff
@@ -76,23 +88,18 @@ func NewConf() (*Conf, error) {
 		rawDiffOutPath = path.Join(rawDiffOutPath, "raw.diff")
 	}
 
-	authEnabled, err := env.GetOrDefault(
```
andream16, line 79: Are we using this anywhere? Perhaps to put the client together

PR #1010, opened 2025-06-13
Add Pagination to Storer interface

ptzianos:


sdk/component/store/local/sqlite/store.go
```diff
@@ -113,21 +116,61 @@ func (m *manager) Validate(*ocsf.VulnerabilityFinding) error {
 
 // Read finds Vulnerability Findings by instanceID.
 // It returns ErrNoFindingsFound is not vulnerabilities were found.
-func (m *manager) Read(ctx context.Context, instanceID uuid.UUID) ([]*vf.VulnerabilityFinding, error) {
-	rawFindings, err := m.queries.FindingsByID(ctx, instanceID.String())
-	if err != nil {
-		return nil, errors.Errorf("could not find findings: %w", err)
+func (m *manager) Read(
```
andream16, line 119: I'd consider always using the paginated method and just using some nice defaults to deal with it in case, so we can remove a lot of code and `FindingsByID`

sdk/component/store/remote/findings-client/client.go
```diff
@@ -182,42 +182,81 @@ func (c *client) Validate(finding *ocsf.VulnerabilityFinding) error {
...
-							ID:      finding.GetId(),
-						}
-					},
-				),
-			), nil
-		default:
-			findingBatches = append(findingBatches, resp.Findings)
-		}
+	findings, err := c.read(ctx, instanceID, queryOpts)
+	findingBatches = append(findingBatches, findings)
```
andream16, line 232: It seems like this can be called one last time after the for loop? 🤔

sdk/component/store/local/sqlite/store.go
ptzianos: that would mean through a lot more extensive reworking of all the components that are calling it. having the query opts set to nil means that the caller doesn't care about how things are fetched but let's the storage layer do its own thing, meaning that it can choose to paginate or not and return the complete results

andream16 (approved):

PR #1016, opened 2025-06-20
Saibamo/add python sdk classes

saibamo:
Finished implementing Python SDK for Enricher.


sdk/python/smithy_python/components/component.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+
+        if not logger:
+            self.log = log
+        elif isinstance(logger, Logger):
+            self.log = logger
+        else:
+            raise TypeError("logger must be an instance of Logger or None.")
+
+        if instance_id is None or (
```
ptzianos, line 42: if the instance_id is a string we should first try to parse it as a uuid4 and if that fails we should raise an exception. the `instance_id` should be a uuid4 field in the class. we can merge the `_validate_uuid` into this part of the constructor. also we should check that it's a UUID4, the Python API doesn't offer a specific identifier so we should do our own checks.

sdk/python/smithy_python/components/component.py
```diff
@@ -0,0 +1,132 @@
...
+                self.db_manager = PostgresDBManager()
+            case DBTypeEnum.REMOTE:
+                self.db_manager = RemoteDBManager(instance_id=self.instance_id)
+            case _:
+                raise ValueError(
+                    "db_type must be one of DBTypeEnum.SQLITE, DBTypeEnum.POSTGRES, or DBTypeEnum.REMOTE."
+                )
+
+    def get_findings(
+        self, page_num: Optional[int] = None
```
ptzianos, line 71: in the Go SDK we allow the user to set the page number and the pagesize and if those are not set, the default ones are used: https://github.com/smithy-security/smithy/blob/main/sdk/component/store/query.go#L5

and 

https://github.com/smithy-security/smithy/blob/main/sdk/component/store/remote/findings-client/client.go#L206

sdk/python/smithy_python/components/component.py
```diff
@@ -0,0 +1,132 @@
...
+
+        :param page_num: Optional parameter to specify the page number for pagination. If not provided, all findings will be retrieved. Otherwise it will retrieve the `SMITHY_REMOTE_CLIENT_PAGE_SIZE`(default 100) findings for the given page number.
+        :type page_num: Optional[int]
+
+        :return: The findings retrieved from the database. This could be a list of `findings_service_pb2.Finding` objects if using the remote database mode, or a list of findings in the format defined by the specific database manager if using `SQLITE` or `POSTGRES`.
+        """
+
+        return self.db_manager.get_findings(page_num=page_num)
+
+    def update_findings(self, findings: List[any]) -> bool:
```
ptzianos, line 85: the findings should be a list of `findings_service_pb2.Finding`

sdk/python/smithy_python/components/component.py
```diff
@@ -0,0 +1,132 @@
...
+            case DBTypeEnum.REMOTE:
+                self.db_manager = RemoteDBManager(instance_id=self.instance_id)
+            case _:
+                raise ValueError(
+                    "db_type must be one of DBTypeEnum.SQLITE, DBTypeEnum.POSTGRES, or DBTypeEnum.REMOTE."
+                )
+
+    def get_findings(
+        self, page_num: Optional[int] = None
+    ) -> Union[List[findings_service_pb2.Finding], List[any]]:
```
ptzianos, line 72: why is the return value a list of `findings_service_pb2.Finding` or `any`? all these calls should use a replica of this type: https://github.com/smithy-security/smithy/blob/main/sdk/component/vulnerability-finding/finding.go#L1

basically the ocsf/v1.VulnerabilityFinding + the finding ID in a data class

ptzianos (commented):
@saibamo please clean up the git history once you have gone through my comments and condense it to fewer commits so that I can do another pass

Makefile
```diff
@@ -63,6 +74,16 @@ test-go: $(go_test_paths)
 cover-go: test-go
 	@go tool cover -html=tests/output/cover.out -o=tests/output/cover.html && open tests/output/cover.html
 
+REVIEWDOG_EXTRA_FLAGS=
```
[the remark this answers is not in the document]
saibamo: done 

https://github.com/smithy-security/smithy/commit/1546fb26c1659f017ddd038c58961dccf340bc65

sdk/python/smithy_python/components/component.py
```diff
@@ -0,0 +1,132 @@
+import uuid
```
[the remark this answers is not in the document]
saibamo: done 

https://github.com/smithy-security/smithy/commit/1546fb26c1659f017ddd038c58961dccf340bc65

sdk/python/smithy_python/components/component.py
```diff
@@ -0,0 +1,132 @@
...
+from smithy_python.helpers.logger import log
```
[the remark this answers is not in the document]
saibamo: Yes! did the refactoring for that!

https://github.com/smithy-security/smithy/commit/1546fb26c1659f017ddd038c58961dccf340bc65

sdk/python/smithy_python/components/component.py
saibamo: If I understand correctly we are (kinda) doing that already, with the exception, that we do not let the user specify the pagesize ad-hoc, but only through the environment vars. But there again, if the Env-var is not set we use the default value.

```
:param page_num: Optional parameter to specify the page number for pagination. If not provided, all findings will be retrieved. Otherwise it will retrieve the `SMITHY_REMOTE_CLIENT_PAGE_SIZE`(default 100) findings for the given page number.
```

Is this something we need to be able to do?

sdk/python/smithy_python/components/enricher.py
```diff
@@ -0,0 +1,55 @@
...
+        :type logger: Optional[Logger]
+        :raises ValueError: If the instance_id is not a valid UUID.
+        :raises TypeError: If the instance_id is not a string or UUID object.
+        :raises NotImplementedError: If the db_type is not `REMOTE`, as currently only `REMOTE` is supported.
+        """
+
+        super().__init__(instance_id, db_type, logger)
+
+    @abstractmethod
+    def enrich(self):
```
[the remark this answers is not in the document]
saibamo: Okay, I see why you only wanted the iterator :D Should I just remove the Enricher Class, or at least encourage ppl in the docs to only use the iteratorcontext?

sdk/python/smithy_python/components/component.py
saibamo: https://github.com/smithy-security/smithy/pull/1016/commits/2d6dd386defa5ba4d664bfba450702486452de38

sdk/python/smithy_python/components/component.py
ptzianos: Ok, there is some duplication in our types that threw me off. We need to eventually clean this up but it's not a problem for now.

To answer your question, yes any DB driver is expected to eventually give us back an object of the protobuf type. How they do this, it's up to the drivers to figure out

sdk/python/smithy_python/components/enricher.py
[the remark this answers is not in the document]
ptzianos: I think we can meet in the happy middle. Lots of things over here will be needed so I don't see a point in removing them, we can just change some names and be ready to go. So why don't we do the following:
- we create a class called `Runner` responsible for parsing environment variables and initialising the DB driver (basically the current Component class) with a method called `run` that accepts an object that's a subclass of the new Component class
- we create a new Component class that's an ABC and an Enricher that's a subclass of that, also an ABC itself with an `enrich` method. the user is expected to subclass the enricher. the `enrich` method gets an iterator and yields the modified findings. if the findings have acquired new enrichments, they are passed to the storage driver to be stored in the db
- we create a storage context manager that resolves the expected DB driver and returns an instance of it that is initialised.
- the `Runner` based on the type of the component, will invoke the storage context manager and will use the storage driver appropriately.

WDYT?

Makefile
```diff
@@ -27,13 +27,26 @@ SMITHY_VERSION=$(shell (echo $(CONTAINER_REPO) | grep -q '^ghcr' && echo $(lates
...
+
 
 export
 
 ########################################
 ######### CODE QUALITY TARGETS #########
 ########################################
-.PHONY: lint install-lint-tools tests test-go fmt fmt-proto fmt-go install-go-fmt-tools
+.PHONY: lint install-lint-tools tests test-go fmt fmt-proto fmt-go install-go-fmt-tools py-tests py-tests-sdk-python update-poetry-pkgs-sdk-python
```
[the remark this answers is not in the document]
saibamo: Yes, added

Makefile
```diff
@@ -27,13 +27,26 @@ SMITHY_VERSION=$(shell (echo $(CONTAINER_REPO) | grep -q '^ghcr' && echo $(lates
...
+install-misspell:
```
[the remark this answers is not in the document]
saibamo: done

sdk/python/smithy_python/components/component.py
saibamo: So do you want me to **OVERRIDE** and/or **OVERWRITE** the env var here?
So I'll add the pape size thing, throw an error when page size but not page num is set.

sdk/python/smithy_python/components/component.py
saibamo: done

sdk/python/smithy/components/runner.py
```diff
@@ -0,0 +1,105 @@
...
+    def _run_enricher(self) -> None:
+        """
+        Runs the logic for the Enricher component.
+
+        This method is responsible for executing the enrichment logic of the component.
+        """
+
+        with self.db_context as db:
+            for finding in db:
+                enriched_finding = self.component.enrich(finding)
```
ptzianos, line 101: we should wrap this invocation in a try except to make sure that we can handle this more nicely

sdk/python/smithy/dbmanagers/remote_db_manager.py
```diff
@@ -0,0 +1,207 @@
...
+            while True:
+                resp = self.stub.GetFindings(
+                    GetFindingsRequest(
+                        id=self.instance_id, page=page, page_size=get_findings_page_size
+                    )
+                )
+                all_findings.extend(resp.findings)
+                if (
+                    len(resp.findings) < get_findings_page_size
+                    or not resp.findings
```
ptzianos, line 112: if `resp.findings` is an empty list, the length will be zero, but if you do `len(None)` then you will get an exception, so I think we should change the order, first if `resp.findings` is not truthy and then the length check

sdk/python/smithy/dbmanagers/resolver.py
```diff
@@ -0,0 +1,79 @@
...
+                )
+            case DBTypeEnum.SQLITE:
+                return SqliteDBManager(instance_id, self._log)
+            case DBTypeEnum.POSTGRES:
+                return PostgresDBManager(instance_id, self._log)
+            case _:
+                raise ValueError(f"Unsupported database type: {db_type}")
+
+    def _load_remote_db_env_vars(self):
+        """
```
ptzianos, line 62: this resolution should be part of the constructor of the remote database client

sdk/python/smithy/dbmanagers/db_context.py
```diff
@@ -0,0 +1,125 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb) -> bool:
+        """
+        Flush any queued updates, then close resources.
+        Exceptions propagate (`return False`) unless you decide otherwise.
+        """
+
+        if self._pending_updates:
+            self._flush_updates()
+
+        if hasattr(self.db_manager, "close"):
```
ptzianos, line 58: this can be simplified by forcing DBManager subclasses to have the `close` method and they just implement a noop if they don't need to have something

sdk/python/smithy/dbmanagers/db_context.py
```diff
@@ -0,0 +1,124 @@
...
+            return next(self._iter)
+
+    def _load_next_page(self) -> None:
+        """
+        Load the next page of findings into the iterator.
+        This method is called when the current page has been exhausted.
+        """
+
+        results = self.db_manager.get_findings(page_num=self._page)
+        self._has_more = bool(results)
```
ptzianos, line 97: if the length of results returned is less than the maximum results per page, we don't need to check the next page, the `_has_more` should be set to `false`

sdk/python/smithy/dbmanagers/db_context.py
```diff
@@ -0,0 +1,124 @@
...
+        self._iter = iter(results)
```
ptzianos, line 98: why use an iterator over here instead of a simple counter that just resets to zero so that we can check how many results in the list are left?

sdk/python/smithy/dbmanagers/db_context.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+
+        if self._pending_updates:
+            self._flush_updates()
+
+        try:
+            self.db_manager.close()
+            self._log.debug("DB connection closed.")
+        except Exception:
+            self._log.exception("Error while closing DB connection")
```
ptzianos, line 62: we are not logging the exception over here that was thrown when closing the connection

sdk/python/smithy/dbmanagers/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .remote_db_manager import RemoteDBManager
```
ptzianos, line 1: why only exposing these classes to the outside? do we need to limit exposure to only these classes?

sdk/python/smithy/components/enricher.py
```diff
@@ -0,0 +1,52 @@
...
+        raise NotImplementedError(
+            "The enrich method must be implemented by subclasses of Enricher."
+        )
+
+        # EXAMPLE
+        enrichment = ocsf_pb2.Enrichment(
+            name="What Kind of Enrichment?",
+            provider="ExampleEnricher",  # Optional
+            type=str(
+                ext_pb2.Enrichment.EnrichmentType.ENRICHMENT_TYPE_UNSPECIFIED
```
ptzianos, line 44: can we add a more realistic scenario over here where the enrichment type is specific and the value is a protojson marshaled object that carries useful information? maybe something closer to the custom annotation enricher?

sdk/python/smithy/dbmanagers/db_context.py
saibamo: I'll postpone this feature request for now.

PR #1019, opened 2025-06-25
Source Code Artifact component

andream16:
Read the [README.md](https://github.com/smithy-security/smithy/blob/a51b624b003b7af10dd7b54b65f7e82233ce69ee/components/targets/source-code-artifact/README.md) for details.

This is a rework of our private s3 target. We are extending it for http archives as well as making it more robust and tested.

This component is able to download and unpack repo archives from:
- s3
- gcs
- plain http(s) urls like https://github.com/andream16/aws-sdk-go-bindings/archive/refs/heads/master.zip

It supports the following file formats:
- zip
- tar
- tar.gz

It essentially:
- configures the artifact backend based on the passed `artifact_url`
- fetches the artifact
- unpacks it based on the file extension into the scratch volume
- outputs the contents in the `sourceCodeWorkspace` for scanners to pick-up
- outputs metadata information

I tried to keep it simple by using dependency injection for the fetchers/extractors so the target itself is very small and highly extensible.

---

How to test:

- **plain http**: run the example workflow `smithyctl workflow run --build-component-images=true --overrides=examples/source-code-artifact/overrides.yaml examples/source-code-artifact/workflow.yaml`
- **gs**: get an example file in a gs bucket like `gs://my-bucket/my-archive.zip`, amend `examples/source-code-artifact/overrides.yaml` parameters, including credentials, and re-run the command above.
- **s3**: same as for gs but using an s3 compatible `archive_url` like `s3://my-bucket/my-archive.tar`


components/targets/source-code-artifact/component.yaml
```diff
@@ -0,0 +1,34 @@
+name: "source-code-artifact"
+description: "Downloads archived files in zip, tag or targz format from a plain http or s3 compatible API"
+type: "target"
+parameters:
+  # Must be a valid url like: https://github.com/aws/aws-sdk-go-v2/archive/refs/heads/main.zip
+  # Or a s3 compatible url like: s3://my-bucket/my-archive.tar or gs://my-bucket/my-archive.tar.gz
+  - name: "artifact_url"
+    type: "string"
+    value: "{{ .context.trigger.event.url }}"
```
ptzianos, line 9: these shouldn't be part of the open source configuration of the components, they are only relevant in the case of smithy saas

components/targets/source-code-artifact/component.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - name: "artifact_reference"
+    type: "string"
+    value: "{{ .context.trigger.event.reference }}"
```
ptzianos, line 12: same thing here

components/targets/source-code-artifact/component.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - name: "artifact_registry_region"
+    type: "string"
+    value: ""
+  - name: "artifact_access_key"
```
ptzianos, line 16: these are weird parameters because in the case of S3 they are the access key ID and secret and in the case of the HTTP auth, username and token. I'm not sure how to name them to make this clear, maybe access_key_or_username? or maybe have completely separate sets of parameters?

components/targets/source-code-artifact/component.yaml
```diff
@@ -0,0 +1,34 @@
+name: "source-code-artifact"
+description: "Downloads archived files in zip, tag or targz format from a plain http or s3 compatible API"
+type: "target"
+parameters:
```
ptzianos, line 4: we should support one more parameter to allow users to set the type of archive explicitly for some artefact that doesn't have a suffix

components/targets/source-code-artifact/go.mod
```diff
@@ -0,0 +1,84 @@
+module github.com/smithy-security/smithy/components/targets/source-code-artifact
+
+go 1.24.2
```
ptzianos, line 3: can we use 1.23.4 instead?

components/targets/source-code-artifact/component.yaml
andream16: Yeah this is not fantastic

What about `auth_id` and `auth_secret`?

senglezou (approved):
LGTM

components/targets/source-code-artifact/component.yaml
andream16: Done

PR #1033, opened 2025-07-14
Configure Smithyctl's docker client's UsePlainHTTP based on --auth-enabled=true flag

andream16:
I think that this is a regression that came out when we extended smithyctl's package capabilities.

This fixes an issue where [server gave HTTP response to HTTPS client](https://stackoverflow.com/questions/49674004/docker-repository-server-gave-http-response-to-https-client) when auth is disabled on packaging.

To test it, you can run:
- `make cmd/smithyctl`
- Spin up https://hub.docker.com/_/registry on port 5000
- `bin/../smithyctl component package --registry-url=127.0.0.1:5000 --sdk-version=v1.0.0 --version=v1.0.0 --registry-auth-enabled=false --registry-plain-http=true components/reporters/json-logger/component.yaml`

### DIFF
--- a/smithyctl/command/component/package.go
+++ b/smithyctl/command/component/package.go
@@ -31,6 +31,7 @@ type (
 		registryAuthEnabled  bool
 		registryAuthUsername string
 		registryAuthPassword string
+		registryPlainHTTP    bool
 		imageRegistryURL     string
 		imageNamespace       string
 		imageTag             string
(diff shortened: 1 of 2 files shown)


senglezou (approved):

smithyctl/command/component/package.go
```diff
@@ -162,15 +162,24 @@ func packageComponent(ctx context.Context, flags packageFlags, componentPath str
...
 	} else {
-		credsStore, err = credentials.NewStoreFromDocker(credentials.StoreOptions{
-			DetectDefaultNativeStore: true,
-		})
+		credsStore, err = credentials.NewStoreFromDocker(
+			credentials.StoreOptions{
+				DetectDefaultNativeStore: true,
+			},
+		)
+		usePlainHTTP = true
```
ptzianos, line 182: shouldn't this be set conditionally based on a flag or something? this looks like it shouldn't be used in production

ptzianos (approved):

PR #1034, opened 2025-07-14
simplify and add reference field to git history enrichment and add remediation enrichment type

ptzianos:


saibamo:
You also need to update the python proto generated classes

andream16 (approved):

PR #1043, opened 2025-07-18
fix: add logging, add correct break conditions for the loop

saibamo:


sdk/python/smithy/dbmanagers/remote_db_manager.py
```diff
@@ -48,50 +50,63 @@ def get_findings(
...
+        )
 
         page = 0
-
         get_findings_page_size = self.page_size
-        if page_num and isinstance(page_num, int) and page_num >= 0:
-            page = page_num
 
-        if page_size and isinstance(page_size, int) and page_size > 0:
+        if page_num is not None and isinstance(page_num, int) and page_num >= 0:
```
ptzianos, line 73: this change is not necessary, a None can never be an instance of an `int`

sdk/python/smithy/dbmanagers/remote_db_manager.py
```diff
@@ -48,50 +50,63 @@ def get_findings(
...
+            page = page_num
+           
+        if page_size is not None and isinstance(page_size, int) and page_size > 0:
```
ptzianos, line 76: same here about the `isinstance` check

ptzianos (approved):

PR #1045, opened 2025-07-18
fix the custom annotation enricher, make it accept variable multiple …

northdpole:
…annotations

### DIFF
--- a/components/enrichers/custom-annotation/component.yaml
+++ b/components/enrichers/custom-annotation/component.yaml
@@ -1,10 +1,14 @@
 name: "custom-annotation"
 description: "Adds a custom annotation enrichment to the passed findings"
 type: "enricher"
+parameters:
+  - name: annotations_key_colon_vals_comma_separated
+    type: string
+    value: ""
 steps:
   - name: "custom-annotation"
     image: "components/enrichers/custom-annotation"
     env_vars:
       CUSTOM_ANNOTATION_NAME: "custom-annotation"
-      CUSTOM_ANNOTATION_VALUES: '{"foo":"bar"}'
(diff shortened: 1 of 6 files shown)


components/enrichers/custom-annotation/component.yaml
```diff
@@ -1,10 +1,14 @@
 name: "custom-annotation"
 description: "Adds a custom annotation enrichment to the passed findings"
 type: "enricher"
+parameters:
+  - name: annotations_key_colon_vals_comma_separated
```
andream16, line 5: This name is a bit overkill imo, I'd rename it to `annotations_key_vals` and document this accordingly

ptzianos:
@northdpole why not expect the user to supply a json encoded map as a parameter? that way we can more easily parse it and clients, especially machine clients can generate this parameter value more easily

northdpole:
> @northdpole why not expect the user to supply a json encoded map as a parameter? that way we can more easily parse it and clients, especially machine clients can generate this parameter value more easily

Valid question, we had several occasions where human users would/could provide raw json in parameters, this wasn't user firendly and caused confusion and errors even when we were using smithy.

- json isn't a format meant for humans to write, hence nobody should have to dump raw json in a field
- you end up in a bind where middleware like frontends or api servers need to encode json multiple times (once for the user value then decode it in the parameter value) and it causes confusion.
- i think a parameter value of: "foo:bar, a:b, c:d" can be shown in frontends easier as tags (esp. with the correct parameter type of map or k-v), if the value type is generic json, we end up in a type confusion of "do we treat it as string or do we somehow prettify how it looks?"
- remember, the parameters are meant for humans, not machines to put together

ptzianos (approved):

PR #1061, opened 2025-08-06
SAAS-7606 Make OSV-Scanner not return an error when input file is emp…

northdpole:
…ty instead of malformed


dlicheva (approved):
Works when I run it locally on a repo without any vulns :raised_hands: also works fine on dvwa.

Tested with ` smithyctl workflow run --build-component-images=true examples/osv-scanner/workflow.yaml` on my repo `https://github.com/dlicheva/secure-repo`

PR #1064, opened 2025-08-14
Added a readme for running simthyctl with the python sdk

PTruscott:


ptzianos:
@PTruscott this readme is for our SaaS repo, we do need to document and make a fix in smithyctl to make sure future users can use the gRPC based storage driver but other than that this README should not be here

PR #1068, opened 2025-08-19
Make transformers not throw an error when receiving an empty resutls file

northdpole:


components/scanners/zaproxy/internal/transformer/transformer_test.go
```diff
@@ -230,4 +231,28 @@ func TestZapTransformer_Transform(t *testing.T) {
 			assert.Equal(t, expectedMetadataUIDs[idx], *finding.Metadata.Uid, "Unexpected metadata uid for finding %d", idx)
 		}
 	})
+	t.Run("it should return an empty finding array when the input file is empty", func(t *testing.T) {
+		emptyFilePath := "./testdata/empty.sarif.json"
```
ptzianos, line 235: we can use a `t.TempDir()` that will be automatically cleaned up when the test case exits, same comment goes for other tests in this PR

components/scanners/zaproxy/internal/transformer/transformer_test.go
```diff
@@ -230,4 +231,28 @@ func TestZapTransformer_Transform(t *testing.T) {
...
+			transformer.ZapRawOutFilePath(emptyFilePath),
+			transformer.ZapTransformerWithTarget(transformer.TargetTypeWebsite),
+			transformer.ZapTransformerWithClock(clock),
+		)
+		require.NoError(t, err)
+
+		ctx := context.WithValue(ctx, component.SCANNER_TARGET_METADATA_CTX_KEY, &ocsffindinginfo.DataSource{
+			TargetType: ocsffindinginfo.DataSource_TARGET_TYPE_WEBSITE,
+		})
+		ctx, cancel := context.WithCancel(ctx)
```
ptzianos, line 251: this context should have a timeout, same goes for other contexts used in these tests

components/scanners/zaproxy/internal/transformer/transformer.go
```diff
@@ -124,7 +124,10 @@ func (g *zapTransformer) Transform(ctx context.Context) ([]*ocsf.VulnerabilityFi
 		}
 		return nil, errors.Errorf("failed to read raw output file '%s': %w", g.rawOutFilePath, err)
 	}
-
+	if len(b) == 0 {
```
ptzianos, line 127: please add empty lines around this block, same comment goes for most of the additions in this PR

components/scanners/sobelow/internal/transformer/transformer_test.go
```diff
@@ -0,0 +1,63 @@
...
+		ctx := context.Background()
+		_, err = transformer.Transform(ctx)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "not found")
+	})
+
+	t.Run("invalid JSON file", func(t *testing.T) {
+		tempFile := filepath.Join(t.TempDir(), "invalid.sarif.json")
+		require.NoError(t, os.WriteFile(tempFile, []byte("invalid json"), 0644))
+		defer func() { require.NoError(t, os.Remove(tempFile)) }()
```
ptzianos, line 53: this is not necessary, it will be cleaned up when the test run stops

components/scanners/codeql/internal/transformer/transformer_test.go
```diff
@@ -213,4 +214,41 @@ func TestCodeQLTransformer_Transform(t *testing.T) {
 			assert.NotNilf(t, affectedCode.StartLine, "Unexpected nil start line for vulnerability for finding %d", idx)
 		}
 	})
+	t.Run("it should return 0 findings when the output file exists but is empty", func(t *testing.T) {
+		// Create an empty SARIF file for testing
+		emptyFilePath := "./testdata/empty_results_sarif.sarif"
```
ptzianos, line 219: this should be written in a `t.TempDir()`

components/scanners/codeql/internal/transformer/transformer_test.go
```diff
@@ -213,4 +214,41 @@ func TestCodeQLTransformer_Transform(t *testing.T) {
...
+		)
+		require.NoError(t, err)
+
+		findings, err := ocsfTransformer.Transform(ctx)
+		require.NoError(t, err)
+		assert.Empty(t, findings, "Expected no findings for an empty SARIF file")
+	})
+	t.Run("it should parse globs with files that are potentially empty but also have files with results", func(t *testing.T) {
+		// Create an empty SARIF file for testing
+		emptyFilePath := "./testdata/empty_results_sarif.json"
```
ptzianos, line 237: this looks like it could (in very rare cases) create some weird issue if the file is not cleaned up correctly with other tests. what if we create a `t.TempDir()` and copy the testdata files into it and create an empty file in it?

PR #1083, opened 2025-08-29
[SAAS-7675] wip - Artefact target - support for apk files

dlicheva:


ptzianos:
we don't need to extract APK files, what we need to do is make sure that that file can be fetched and moved to the correct location in the filesystem. will open another PR to tackle this

PR #1089, opened 2025-09-10
Use relative paths instead of absolute

senglezou:


components/scanners/trufflehog/internal/transformer/transformer.go
```diff
@@ -298,12 +306,20 @@ func (t *trufflehogTransformer) mapDataSource(ctx context.Context, location Truf
 	if location.SourceMetadata.Data.Filesystem.File == "" {
 		return "", errors.Errorf("unsupported trufflehog findings with empty file entry detected, finding %#v", location)
 	}
+
+	fullPath := location.SourceMetadata.Data.Filesystem.File
+	relativePath, err := filepath.Rel(t.stripFilePathPrefix, fullPath)
```
senglezou, line 311: We could use relativePath, err := strings.TrimPrefix(fullPath, t.stripFilePathPrefix) instead?

components/scanners/trufflehog/internal/transformer/transformer.go
```diff
@@ -171,10 +172,17 @@ func New(opts ...TrufflehogTransformerOption) (*trufflehogTransformer, error) {
 		return nil, err
 	}
 
+	stripFilePathPrefix, _ := env.GetOrDefault(
+		"TRUFFLEHOG_SOURCE_CODE_WORKSPACE",
+		"/workspace/source-code",
```
ptzianos, line 177: I don't think we should have a default value over here, instead it should fail if this value is not provided

components/scanners/trufflehog/internal/transformer/transformer.go
ptzianos: I think it looks fine as is, better to use a function that is made for unix paths than a plain string manipulation one

components/scanners/trufflehog/internal/transformer/transformer.go
```diff
@@ -298,12 +306,20 @@ func (t *trufflehogTransformer) mapDataSource(ctx context.Context, location Truf
 	if location.SourceMetadata.Data.Filesystem.File == "" {
 		return "", errors.Errorf("unsupported trufflehog findings with empty file entry detected, finding %#v", location)
 	}
+
+	fullPath := location.SourceMetadata.Data.Filesystem.File
+	relativePath, err := filepath.Rel(t.stripFilePathPrefix, fullPath)
+	if err != nil {
+		// If there's an error, fall back to the full path to avoid losing data
+		relativePath = fullPath
```
ptzianos, line 314: if there is an error over here we should not proceed, it means that there is probably a misconfiguration

components/scanners/trufflehog/internal/transformer/transformer.go
```diff
@@ -298,12 +313,19 @@ func (t *trufflehogTransformer) mapDataSource(ctx context.Context, location Truf
...
+		return "", errors.Errorf("absolute path %q does not have expected prefix %q", fullPath, t.stripFilePathPrefix)
+	}
+	relativePath := strings.TrimPrefix(fullPath, t.stripFilePathPrefix)
+
 	targetMetadata := component.TargetMetadataFromCtx(ctx)
 	dataSource := ocsffindinginfo.DataSource{
 		TargetType: t.targetType,
 		Uri: &ocsffindinginfo.DataSource_URI{
 			UriSchema: ocsffindinginfo.DataSource_URI_SCHEMA_FILE,
-			Path:      location.SourceMetadata.Data.Filesystem.File,
```
senglezou, line 306: does this need to be `Path: utils.Ptr(fmt.Sprintf("file://%s", relativePath)),` like the affectedCode block?? I don't have context why we didn't add the file:// bit from the beginning @northdpole 

thank you 🙏🏻

PR #1125, opened 2025-11-19
add check for product uid in OCSF validation helpers

ptzianos:


PTruscott (approved):

PTruscott (approved):
Is it worth adding a test for the new affectedpackage section?  That if it exists, it's not malformed?

ptzianos:
> Is it worth adding a test for the new affectedpackage section? That if it exists, it's not malformed?

yes, but at the moment we are also not sure what a valid affected package looks like. I think we should discuss separately that subject when we start using it for deduplication or some sort of post-processing

